PR #138, opened 2014-09-11
add read_file, write_file, mkdir, adjust_permission functions in vsc.utils.missing

boegel:


test/filetools.py
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python
+##
+#
+# Copyright 2012-2013 Ghent University
```
JensTimmerman: 2015

lib/vsc/utils/filetools.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def read_file(path, log_error=True):
+    """Read contents of file at given path, in a robust way."""
+    f = None
+    # note: we can't use try-except-finally, because Python 2.4 doesn't support it as a single block
+    try:
+        f = open(path, 'r')
+        txt = f.read()
+        f.close()
```
JensTimmerman: it's never bad to explicitly call close.

lib/vsc/utils/filetools.py
```diff
@@ -0,0 +1,201 @@
...
+            except OSError, err:
+                _log.raiseException("Failed to set groud ID/sticky bit: %s" % err)
+    else:
+        _log.debug("Not creating existing path %s" % path)
+
+
+def read_file(path, log_error=True):
+    """Read contents of file at given path, in a robust way."""
+    f = None
+    # note: we can't use try-except-finally, because Python 2.4 doesn't support it as a single block
```
JensTimmerman: no need to support python 2.4 anymore, YAY!

lib/vsc/utils/filetools.py
```diff
@@ -0,0 +1,201 @@
...
+        f.close()
+        return txt
+    except IOError, err:
+        # make sure file handle is always closed
+        if f is not None:
+            f.close()
+        if log_error:
+            _log.raiseException("Failed to read %s: %s" % (path, err))
+        else:
+            return None
```
JensTimmerman: No need to `return None` when an exceptional case occurs, thrown an exception instead, that's what they're for. 
Flow control improves when using exceptions instead of having to check return values all the time.

PR #142, opened 2014-09-15
make existing code Py3 compatible (WIP)

boegel:
This effectively makes `vsc-base` require Python 2.6, so be careful when merging this in (EasyBuild, which needs `vsc-base` will only move to Python 2.6 early 2015).

Currently WIP, not all unit tests pass yet with Python3 (but they still do using Python2.6).

Maybe this should include a bump to v2.0?

### DIFF
--- a/lib/vsc/utils/asyncprocess.py
+++ b/lib/vsc/utils/asyncprocess.py
@@ -107,8 +107,10 @@ class Popen(subprocess.Popen):
             return 0
 
         try:
+            if isinstance(inp, str):
+                inp = inp.encode()
             written = os.write(self.stdin.fileno(), inp)
-        except OSError, why:
+        except OSError as why:
             if why[0] == errno.EPIPE:  # broken pipe
                 return self._close('stdin')
             raise
@@ -177,7 +179,7 @@ def recv_some(p, t=.1, e=False, tr=5, stderr=False, maxread=None):
             len_y += len(r)
         else:
             time.sleep(max((x - time.time()) / tr, 0))
-    return ''.join(y)
+    return b''.join(y)
 
 
 def send_all(p, data):
@@ -188,4 +190,9 @@ def send_all(p, data):
         sent = p.send(data)
         if sent is None:
             raise Exception(MESSAGE)
-        data = buffer(data, sent)
+        try:
+            if isinstance(data, str):
+                data = data.encode()
+            data = memoryview(data) #[:sent]  # sent?
+        except NameError:
+            data = buffer(data, sent)
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -72,6 +72,9 @@ Logging to a udp server:
 @author: Kenneth Hoste (Ghent University)
 """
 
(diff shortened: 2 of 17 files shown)


boegel:
I still care about this, should I revive it and fix the remark,s @stdweird?

I hope to have Python 3 support in EB v3.

stdweird:
we should wait after jens his talk.
@JensTimmerman you can look at this PR and add some examples how to move during the presentation?

boegel:
OK, on hold until further notice then.

There are a couple of more involved things I couldn't figure out, so this PR is certainly not complete, but it a major step forward towards Py3 compatibility.

lib/vsc/utils/asyncprocess.py
```diff
@@ -188,4 +190,9 @@ def send_all(p, data):
         sent = p.send(data)
         if sent is None:
             raise Exception(MESSAGE)
-        data = buffer(data, sent)
+        try:
+            if isinstance(data, str):
+                data = data.encode()
+            data = memoryview(data) #[:sent]  # sent?
+        except NameError:
+            data = buffer(data, sent)
```
JensTimmerman: is this really what you intended?
`data` could be encoded here, or maybe not.

lib/vsc/utils/fancylogger.py
```diff
@@ -72,6 +72,9 @@
 @author: Kenneth Hoste (Ghent University)
 """
 
+# required to avoid Py2 interpreting "print('a', 'b')" as printing a tuple
```
JensTimmerman: you want to reimplement six ?

lib/vsc/utils/frozendict.py
```diff
@@ -17,14 +17,20 @@
...
-from UserDict import DictMixin
-
+import sys
+from distutils.version import LooseVersion
+pyver = sys.version.split(' ')[0]
+if LooseVersion(pyver) < LooseVersion('3.0'):
+    from UserDict import DictMixin as MutableMapping
+else:
+    # see http://stackoverflow.com/questions/11165188/how-to-achieve-the-functionality-of-userdict-dictmixin-in-python-3
+    from collections import MutableMapping
```
JensTimmerman: this ugly check shouldn't be here at all, collections.MutableMapping is available in python 2.6, yes, everything is easier when you drop python 2.4 and 2.5 ;-)

lib/vsc/utils/generaloption.py
```diff
@@ -50,6 +48,13 @@
 from vsc.utils.missing import shell_quote, nub
 from vsc.utils.optcomplete import autocomplete, CompleterOption
 
+pyver = sys.version.split(' ')[0]
+if LooseVersion(pyver) < LooseVersion('3.0'):
+    from ConfigParser import ConfigParser
+    from StringIO import StringIO
+else:
+    from configparser import ConfigParser
+    from io import StringIO
```
JensTimmerman: `from io import StringIO` works fine in python 2.6

boegel:
@JensTimmerman you were looking into an approach with `six`?

boegel:
follow-up in #231

PR #151, opened 2015-01-13
enhance shell_quote to wrap token in single quotes when needed

boegel:
I ran into this when configuring EasyBuild with a custom log format (`--log-format='easybuild,eb-%(name)s.log'`) and then using `--job`, which uses `generate_cmd_line` to dump the full configuration as a list of command line options, which uses `shell_quote` internally.

Without this, the value passed to `--log-format` is included unquoted (`--log-format=easybuild,eb-%(name)s.log`) in the job command line, which then fails with `syntax error near unexpected token `('`.

@stdweird: please review


lib/vsc/utils/missing.py
```diff
@@ -282,11 +281,20 @@ def __getitem__(self, key, *args, **kwargs):
...
+    """
+    # make sure we're dealing with a string
+    if not isinstance(token, basestring):
+        token = str(token)
+    # check whether token is already wrapped in single quotes
+    is_escaped = re.compile("^'.*'$")
+    if not is_escaped.match(token):
+        # escape any non-escaped single quotes, and wrap entire token in single quotes
+        token = "'%s'" % re.sub(r"([^\\])'", r"\1\\'", token)
+    return token
```
JensTimmerman: do you really want to write your own quoter? this looks like a bug in list2cmdline?

boegel: `list2cmdline` doesn't take special characters like `$`, `)`, `(` into account at all, it only cares about whitespace and backslashes; besides, it's a (purposely) undocumented function, so better not to rely on it...

simply quoting everything with single quotes (carefully) is better/safer imho... we want to avoid that the shell does any interpretation at all

stdweird: you need to test for 
```python
"''''
```
sp the whole function becomes
```python
return "'%s'" % re.sub(r"([^\\])'", r"\1\\'", str(token))
```

but i'll have a look what corener cases we are missing here that are covered by `list2cmdline`

lib/vsc/utils/missing.py
```diff
@@ -282,11 +281,20 @@ def __getitem__(self, key, *args, **kwargs):
...
 
-def shell_quote(x):
-    """Add quotes so it can be apssed to shell"""
-    # use undocumented subprocess API call to quote whitespace (executed with Popen(shell=True))
-    # (see http://stackoverflow.com/questions/4748344/whats-the-reverse-of-shlex-split for alternatives if needed)
-    return subprocess.list2cmdline([str(x)])
+def shell_quote(token):
+    """
+    Wrap provided token in single quotes (to escape space and characters with special meaning in a shell),
+    so it can be used in a shell command.
```
stdweird: speciy that this function does not allow shell expansion/inerpolation

test/missing.py
```diff
@@ -367,6 +367,26 @@ def test_avail_subclasses_in(self):
         # cleanup
         sys.path = orig_sys_path
 
+    def test_shell_quote(self):
+        """Test shell_quote function"""
+        values = [
+            (123, "'123'"),
```
stdweird: and add a test for string with escaped single quotes 
```python
a="\'"
```

lib/vsc/utils/missing.py
```diff
@@ -282,11 +281,20 @@ def __getitem__(self, key, *args, **kwargs):
...
+    """
+    # make sure we're dealing with a string
+    if not isinstance(token, basestring):
+        token = str(token)
+    # check whether token is already wrapped in single quotes
+    is_escaped = re.compile("^'.*'$")
+    if not is_escaped.match(token):
+        # escape any non-escaped single quotes, and wrap entire token in single quotes
+        token = "'%s'" % re.sub(r"([^\\])'", r"\1\\'", token)
```
stdweird: you need negatove look-ahead for this, but it's unclear what you are trying here. make sure to add tests for it.

test/missing.py
```diff
@@ -367,6 +367,30 @@ def test_avail_subclasses_in(self):
...
+            # shell_unqoute can't unquote a single-quoted string with (single) quotes in it (ValueError: No closing quotation)
+            ("foo'bar\\'baz", "'foo\\'bar\\'baz'", False),
+            ("'", "'\\''", False),
+            ("\'", "'\\''", False),
+            ("\\'", "'\\''", False),
+            ("''", "'\\'\\''", False),
+        ]
+        for orig_value, quoted_value, test_unquote in values:
+            self.assertEqual(shell_quote(orig_value), quoted_value)
+            if test_unquote:
```
stdweird: sorry, can't have this. this has to be tested

test/missing.py
```diff
@@ -367,6 +367,30 @@ def test_avail_subclasses_in(self):
...
+            (123, "'123'", True),
+            ('foo', "'foo'", True),
+            ('value with whitespace', "'value with whitespace'", True),
+            ('foo\tbar', "'foo\tbar'", True),
+            ('(value)', "'(value)'", True),
+            ('$value', "'$value'", True),
+            ('value with (foo)', "'value with (foo)'", True),
+            ('value with $foo', "'value with $foo'", True),
+            # check that escaped single quotes aren't escaped again
+            # shell_unqoute can't unquote a single-quoted string with (single) quotes in it (ValueError: No closing quotation)
```
JensTimmerman: if it can't do it then it is broken and should be fixed?

test/missing.py
```diff
@@ -367,6 +367,30 @@ def test_avail_subclasses_in(self):
...
+            ("'", "'\\''", False),
+            ("\'", "'\\''", False),
+            ("\\'", "'\\''", False),
+            ("''", "'\\'\\''", False),
+        ]
+        for orig_value, quoted_value, test_unquote in values:
+            self.assertEqual(shell_quote(orig_value), quoted_value)
+            if test_unquote:
+                self.assertEqual(str(orig_value), shell_unquote(shell_quote(orig_value)))
+
```
JensTimmerman: if you're sure you want the value error to be ok, then you have to test for it, so add an else case and assert that it raises the error.

geimer:
@boegel asked me to ping this PR.  Any chance that it gets revisited again anytime soon?  It's hitting me as well (see the EB framework issue linked above).

stdweird:
@geimer @boegel this PR is closed, not sure what this is about?

geimer:
@stdweird: This PR was closed w/o being merged for whatever reason. However, I'm also experiencing issues with `shell_quote` in vsc-base being broken (see https://github.com/hpcugent/easybuild-framework/issues/1438), which is addressed here.  This PR fixes the problem for me and I therefore believe that it should be investigated again.  Does this answer your question?

stdweird:
@geimer apologies, i thought it was merged instead of closed. i'll try to have a look

boegel:
reopening this, it may need some work to get it in, but it seems like a proper fix for the issue @geimer is running into in EasyBuild

stdweird:
@boegel i don't think this is the right approach to the underlying problem (or a bug that causes part of it).

boegel:
@stdweird: why not? Do you have another/better solution in mind? It actually fixes the problem at hand.

stdweird:
the problem is that generaloption uses `shell_quote` to generate the commandline, while it needs a serialiser of some sort (like what you made out of `shell_quote`). also, the unittests should test this by invoking an actual shell and a command to verify the generated commandline).

boegel:
@wpoely86 you have a good idea for the proper way to fix this?

JensTimmerman:
can you explain why this was closed? and how it is fixed now?

wpoely86:
Yes, this is not the proper way. 
As @stdweird said, we should have some sort of serializer so that options -> generate_cmdline -> options works (meaning that both options are equivalent).

boegel:
@JensTimmerman also tracked via https://github.com/hpcugent/vsc-base/issues/152 and https://github.com/hpcugent/easybuild-framework/issues/1438, so we won't forget about this issue easily

PR #153, opened 2015-02-07
define vsc-base version in a dedicated vsc.version module, for easy access

boegel:
I need this so I can check the version of the vsc-base installation found by EasyBuild, since it may not have been installed via `easy_install`/`pip`

@stdweird: please review?


stdweird:
i'm not sure the namespace is correct. shouldn't `vsc-base` rule all things `vsc.utils` (and thus have a `vsc.utils.VERSION`? @JensTimmerman thougths?

boegel:
Then it should be called vsc-utils? And isn't there a package with that name already? :-)

I'm happy to move it somewhere else, but this vsc.version makes sense to me, since all other vsc packages depend on vsc-base anyway...

boegel:
I can also move the definition of VERSION into vsc/__init__.py

JensTimmerman:
https://www.python.org/dev/peps/pep-0008/#version-bookkeeping

You should consider using this 'template' and let vsc.install.setup set it to the correct value during installation (the version that's in the setup.py file.

this way every installed file is versioned
and 

```
$ python
>>> from vsc import utils
>>> utils.__version__
```
will work like it works for a lot of packages

boegel:
Except when `vsc-base` is not being installed via `setup.py` (e.g. unpack + modify `$PYTHONPATH`)...

stdweird:
you should only expect vsc-base to work with `setup.py install`, everything else, well, they should know what they are doing.

boegel:
OK, will, in that case, I don't really need what's here... Either I have a way of checking the version of `vsc-base` that works in all situations, or I don't check the version (or, leave it up to `setup.py` to make sure a compatible version of `vsc-base` is around).

See https://github.com/hpcugent/easybuild-framework/pull/1160/files#diff-3a81198e36d04fed032676148168cb68R44

Unless there are any objections, I'll just close this, and use `vsc-base` v2.0.0 as a dependency for EasyBuild v2.0.0 (omg, they match up! :P).

boegel:
I want to see this reconsidered... Since the release of EasyBuild v2.0, which uses vsc-base as a proper dependency, people have been running into confusing errors after they updated their EasyBuild framework copy without updating vsc-base too.

Having the vsc-base version available in a constant, that the EasyBuild framework can check, would allow to emit a very clear error message like:

```
ERROR: Found vsc-base v2.1.3, but v2.2.0 (or newer) is required
```

Since this is a very light-weight check, and we can import the version into setup.py rather than hardcoding it there (too), I see no reason not to go through with this.

@stdweird, @JensTimmerman: thoughts?

stdweird:
i agre on the cleaner error message, and we can reconsider this, but certainly not in this form. it's simplistic at best and wrong since vsc-base is about `vsc.utils`. i want to know if there are better ways to do this (like get the version via setuptools; although i don't know how eg rpms would handle that)

i'm also afraid that to do this right, we might need versions per module (not a big issue per se). can we do templating via setup.py?

wrt EB, lets hope that any other dependecies play by the same rules or also provide version controls.

JensTimmerman:
Adding a vsc.version will not work either, who governs this namespace? You cant have vsc.version for every `vsc-` package, since they would conflict. 

So I propose to use the `__version__`, as proposed by the Python community. 
If you are willing to cater to people who don't install but just unpack, you have the option of checking the existance of the `setup.py` file, and import the version from there, when the `__version__` fails to provide results. (Since the setup.py file will be present after just unpacking)

boegel:
I'll look into @JensTimmerman's suggestion, at some point...

PR #154, opened 2015-02-10
python2 should be the interpreter for binary files

JensTimmerman:
libraries should not have a shebang.

### DIFF
--- a/bin/logdaemon.py
+++ b/bin/logdaemon.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 #
 # Copyright 2011-2013 Ghent University
 #
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 # -*- coding: latin-1 -*-
 #
 # Copyright 2009-2014 Ghent University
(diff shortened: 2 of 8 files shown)


bin/logdaemon.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
```
boegel: `python2` isn't there everywhere, so we need to be more careful with this somehow, maybe...

I don't have `python2` on OS X, for example...

setup.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
```
boegel: I'm not sure whether a shebang like this even makes sense here, since you're always running `python setup.py install` anyway?

JensTimmerman: no, you can run `./setup.py install`

bin/logdaemon.py
JensTimmerman: ok, then I got this wrong, I believed a python2 binary was always present since python 2.6
```

easybuild-framework (system_config) $ which python
/Library/Frameworks/Python.framework/Versions/2.7/bin/python
easybuild-framework (system_config) $ which python2
/Library/Frameworks/Python.framework/Versions/2.7/bin/python2
```

JensTimmerman:
not relevant

PR #155, opened 2015-02-16
stop using deprecated 'message' attribute

boegel:
EasyBuild users sometimes run into this:

```
<prefix>vsc/utils/generaloption.py:1031: DeprecationWarning: BaseException.message has been deprecated as of Python 2.6
```

But there's really no reason to use `err.message anymore:

Python 2.6:
```
$ python
Python 2.6.6 (r266:84292, Jan 22 2014, 05:06:49) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-3)] on linux2
>>> class A():
...  pass
... 
>>> a=A()
>>> try:a.foo
... except AttributeError, err: print str(err)
... 
A instance has no attribute 'foo'
```

Python 2.7:
```
$ python
Python 2.7.6 (default, Sep  9 2014, 15:04:36) 
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.39)] on darwin
>>> class A():
...   pass
... 
>>> a=A()
>>> try:a.foo
... except AttributeError, err: print err.message
... 
A instance has no attribute 'foo'
>>> try:a.foo
... except AttributeError, err: print str(err)
... 
A instance has no attribute 'foo'
```

### DIFF
--- a/lib/vsc/utils/generaloption.py
+++ b/lib/vsc/utils/generaloption.py
@@ -1027,7 +1027,7 @@ class GeneralOption(object):
         try:
             (self.options, self.args) = self.parser.parse_args(options_list)
         except SystemExit, err:
-            self.log.debug("parseoptions: parse_args err %s code %s" % (msg, err.code))
+            self.log.debug("parseoptions: parse_args err %s code %s" % (err, err.code))
             if self.no_system_exit:
                 return
             else:
--- a/lib/vsc/utils/generaloption.py
+++ b/lib/vsc/utils/generaloption.py
@@ -1027,11 +1027,6 @@ class GeneralOption(object):
         try:
             (self.options, self.args) = self.parser.parse_args(options_list)
         except SystemExit, err:
-            try:
-                msg = err.message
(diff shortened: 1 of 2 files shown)


boegel:
@stdweird: please review

stdweird:
@boegel how do you trigger the deprecation warning? i don't get it in 2.7.5.

wrt the PR, it's not the same so i'll have to understand what is different.

boegel:
@stdweird: you can trigger the issue using `eb --version` with Python 2.6 (I'm not seeing it with Python 2.7 either)

the change just uses `str(err)` basically, since it's being used to complete a template

the code I copy-pasted in the description shows that this works fine with Python 2.6 and 2.7, and that the string is 'clean', i.e. no `AttributeError` in the value of `str(err)`

boegel:
@stdweird: ping on this?

stdweird:
OK for me to go in.

boegel:
Thanks @stdweird!

PR #157, opened 2015-02-20
add support for (un)setting options back to None in configuration files

boegel:
I need this in order to make the EasyBuild framework unit tests robust against any user configuration files that are present (which will be part of https://github.com/hpcugent/easybuild-framework/pull/1167)

### DIFF
--- a/lib/vsc/utils/generaloption.py
+++ b/lib/vsc/utils/generaloption.py
@@ -30,6 +30,7 @@ A class that can be used to generated options to python scripts in a general way
 
 @author: Stijn De Weirdt (Ghent University)
 @author: Jens Timmerman (Ghent University)
+@author: Kenneth Hoste (Ghent University)
 """
 
 import ConfigParser
@@ -41,8 +42,9 @@ import re
 import StringIO
 import sys
 import textwrap
-from optparse import OptionParser, OptionGroup, Option, Values, HelpFormatter
-from optparse import BadOptionError, SUPPRESS_USAGE, NO_DEFAULT, OptionValueError
+from distutils.version import LooseVersion
+from optparse import OptionParser, OptionGroup, Option, Values
+from optparse import BadOptionError, SUPPRESS_USAGE, OptionValueError
 from optparse import SUPPRESS_HELP as nohelp  # supported in optparse of python v2.4
 from optparse import _ as _gettext  # this is gettext normally
 from vsc.utils.dateandtime import date_parser, datetime_parser
@@ -1058,7 +1060,16 @@ class GeneralOption(object):
                 the 2nd level key,value is the key=value. 
                 All section names, keys and values are converted to strings.
         """
-        self.configfile_parser = self.CONFIGFILE_PARSER()
+        # allow_no_value indicates that options can be unset (a.k.a. set to None) in configuration files,
+        # by specifying them without a value (for example: 'foo' rather than 'foo = bar')
+        # note: only available with Python 2.7
+        # see also https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser
+        if LooseVersion(sys.version) >= LooseVersion('2.7'):
+            self.log.debug("Initializing configfile parser, enable allowing to unset options")
+            self.configfile_parser = self.CONFIGFILE_PARSER(allow_no_value=True)
+        else:
+            self.log.debug("Initializing configfile parser, unsetting of options is not supported < Py2.7")
+            self.configfile_parser = self.CONFIGFILE_PARSER()
 
         # make case sensitive
         if self.CONFIGFILE_CASESENSITIVE:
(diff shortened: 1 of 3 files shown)


boegel:
@stdweird, @JensTimmerman: please review

lib/vsc/utils/generaloption.py
```diff
@@ -1058,7 +1060,16 @@ def configfile_parser_init(self, initenv=None):
...
-        self.configfile_parser = self.CONFIGFILE_PARSER()
+        # allow_no_value indicates that options can be unset (a.k.a. set to None) in configuration files,
+        # by specifying them without a value (for example: 'foo' rather than 'foo = bar')
+        # note: only available with Python 2.7
+        # see also https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser
+        if LooseVersion(sys.version) >= LooseVersion('2.7'):
+            self.log.debug("Initializing configfile parser, enable allowing to unset options")
+            self.configfile_parser = self.CONFIGFILE_PARSER(allow_no_value=True)
+        else:
+            self.log.debug("Initializing configfile parser, unsetting of options is not supported < Py2.7")
```
JensTimmerman: well if the config file can have undefined behaviour this should be warning ?

boegel: behaviour won't be undefined, if you try to unset options using Python 2.6, you'll run into a `ParserError`, see unit tests

I'm happy to change this to 'warning' though

stdweird:
i don't like this, because it's different behaviour for 2.6 vs 2.7
what's the usecase?

boegel:
@stdweird: I don't like that part either, but at least it'll work with Python 2.7

I need this to make the EB unit tests more robust. I'm looking into generating a configuration file that sets all the configuration options to their default value, and installing that before starting the unit test suite. My goal is to make the tests run with the default configuration options, regardless of which system or user configuration files are available on the system.

Setting configuration options that have a non-`None` default value works, but making sure that the others are unset (i.e., back to `None`) is not possible without this, hence configuration options like `robot` can not be reset to `None` to make the unit tests robust against custom configuration...

stdweird:
if you want to ignore any config file, just use `--configfiles=/dev/null`?

boegel:
@stdweird: since the action type for `configfiles` is `add`, that won't make a difference? won't the default config file always be picked up?

stdweird:
ofcourse. then either set `CONFIGFILES_USE` to False (monkey patch it) or add the default files to the `--ignoreconfigfiles`.

boegel:
ah, yes, that would be tons easier... I can't set `CONFIGFILES_USE` to `False`, but I can use `--ignoreconfigfiles` (or rather, `$EASYBUILD_IGNORECONFIGFILES`) to make sure any existing configuration files are being ignored

nevertheless, this feature of allowing to unset `None`-default options may be useful still?

stdweird:
why can't you set `CONFIGFILES_USE` to `False`? just a class constant.

boegel:
because some of the unit tests use (temporary) configfiles themselves

and I'm not in favour of global monkey-patching stuff in the tests, it will hide problems sooner or later

stdweird:
it's only in the unittests. nothing global about it (you can restore it in the cleanup part of the test. if it works, it's a lot cleaner then using `ignoreconfigfiles`

boegel:
how can monkey patching be cleaner than defining `$EASYBUILD_IGNORECONFIGFILES`? see https://github.com/boegel/easybuild-framework/blob/2545e3d31bd00a764fe2d6afe5ef23f0c26f8832/test/framework/utilities.py#L72

besides that, I would have to undo the monkey patching in several places, i.e. where I want to test the behaviour with config files being provided

boegel:
I don't need this anymore for https://github.com/hpcugent/easybuild-framework/pull/1167 since using `EASYBUILD_IGNORECONFIGFILES` is a superior approach to putting a config file in place that tries to reset everything to default

but I still feel unsetting/resetting options in a config file is useful to support, despite the fact that it'll only work with Py2.7 (trying to use a config file that resets values in Py2.6 will results in a `ParserError` being thrown)

stdweird:
if we ever really need this, we might reconsider. still don't know a valid usecase though ...

boegel:
The use case is resetting a configuration option back to `None`, after it was set by another configuration option which was processed earlier.

Say there are two (EasyBuild) configuration files listed with `--configfiles=/tmp/one.cfg,/tmp/two.cfg`.

`one.cfg`:
```
[basic]
# enable dependency resolution, use easyconfigs in /tmp
robot = /tmp
```

`two.cfg`:
```
[basic]
# do not enable robot by default, (re)set to None
robot
```

Without this patch resetting a previously set configuration option back to `None` is not possible.
`two.cfg` will fail to parse because `allow_no_value` was not set to `True` when initializing `ConfigParser`, resulting in a `ParserError` being thrown.

stdweird:
with this patch, the example doesn't work in 2.6, so lots of nice surprises for users.
and with usecase i meant actual usecase. there are ways around this (purely academic) example, so (still) no reason to merge this in.

boegel:
It doesn't work with Py2.6 no, but that's not something we can change.

How would you, as an end user, reset a configuration option that was defined  in a system configuration file back to `None` (without this in place)? I don't consider that an academic example.

Anyone, I don't care enough about this to keep fighting over this since I no longer have an urgent need for this, I just feel this would be nice to have.

So, shall I just close this?

JensTimmerman:
I guess this issue arrises from having the weird Robot option, which has a double meaning? The True or store setting can give this problem, but couldn't you just set it to False?

stdweird:
yeah, i thought about that. support for `store_or_None` in a config file would be a use case, but i was waiting for an actual one. and the issue of no 2.6 support remains. i'd rather say doesn't work then works on python X

PR #158, opened 2015-02-20
fix stupid typo in debug log message when --ignoreconfigfiles is used

boegel:


lib/vsc/utils/generaloption.py
```diff
@@ -1116,7 +1116,7 @@ def parseconfigfiles(self):
                     self.log.debug("parseconfigfiles: configfile %s not found, will be skipped" % fn)
 
             if fn in option_ignoreconfigfiles:
-                self.log.debug("parseconfigfiles: configfile %s will be ignored %s" % fn)
+                self.log.debug("parseconfigfiles: configfile %s will be ignored" % fn)
```
JensTimmerman: beter put str(fn) here, looks like it can be a tuple sometimes?

JensTimmerman: or actually
`self.log.debug("parseconfigfiles: configfile %s will be ignored",fn)`

PR #159, opened 2015-03-09
fix fallback mechanism to distutils for bdist_rpm in shared_setup.py

boegel:
there's no `_bdist_rpm` in `distutils`...

this should have been tackled along with #133 (maybe), but it wasn't


lib/vsc/install/shared_setup.py
```diff
@@ -104,12 +104,14 @@ def find_sources(self):
                 self.filelist.files.remove(f)
 
     has_setuptools = True
-except ImportError:
+except ImportError, err:
```
JensTimmerman: Please use `except ImportError as err:` it's clearer.

PR #160, opened 2015-03-11
implement LoggedException in vsc.utils.exceptions module

boegel:
implement idea for self-logging exception mentioned by @riccardomurri in https://github.com/hpcugent/easybuild-framework/pull/1183

### DIFF
--- a/lib/vsc/utils/exceptions.py
+++ b/lib/vsc/utils/exceptions.py
@@ -67,15 +67,14 @@ class LoggedException(Exception):
     # logger module to use (must provide getLogger() function)
     LOGGER_MODULE = fancylogger
     # name of logging method to use (must accept a single argument of type string, i.e. the log message)
-    LOGGING_METHOD = 'error'
+    # note: 'self' is passed as 1st argument, which is just ignored
+    LOGGING_METHOD = lambda _, logger, msg: logger.error(msg)
 
     def __init__(self, msg, logger=None):
         """
         Constructor.
         @param logger: logger to use
         """
-        super(LoggedException, self).__init__(msg)
-
         # try to use logger defined in caller's environment
         if logger is None:
             logger = get_callers_logger()
@@ -83,4 +82,6 @@ class LoggedException(Exception):
             if logger is None:
                 logger = self.LOGGER_MODULE.getLogger()
 
-        getattr(logger, self.LOGGING_METHOD)(msg)
+        self.LOGGING_METHOD(logger, msg)
+
+        super(LoggedException, self).__init__(msg)
--- a/lib/vsc/utils/exceptions.py
+++ b/lib/vsc/utils/exceptions.py
@@ -44,7 +44,11 @@ def get_callers_logger():
     logger = None
     try:
         # frame may be None, see https://docs.python.org/2/library/inspect.html#inspect.currentframe
-        framerecords = (frame is None and []) or inspect.getouterframes(frame)
+        if frame is None:
+            framerecords = []
+        else:
+            framerecords = inspect.getouterframes(frame)
+
         for frameinfo in framerecords:
             bindings = inspect.getargvalues(frameinfo[0]).locals
             for val in bindings.values():
--- a/lib/vsc/utils/exceptions.py
+++ b/lib/vsc/utils/exceptions.py
@@ -66,8 +66,9 @@ class LoggedException(Exception):
 
     # logger module to use (must provide getLogger() function)
     LOGGER_MODULE = fancylogger
-    # name of logging method to use (must accept a single argument of type string, i.e. the log message)
-    # note: 'self' is passed as 1st argument, which is just ignored
+    # logging method to use
+    # takes a logger instance and an argument of type string, i.e. the log message
+    # note: 'self' (LoggedException instance) is passed as 1st argument, which is typically just ignored
     LOGGING_METHOD = lambda _, logger, msg: logger.error(msg)
 
     def __init__(self, msg, logger=None):
--- a/lib/vsc/utils/exceptions.py
+++ b/lib/vsc/utils/exceptions.py
@@ -42,28 +42,32 @@ def get_callers_logger():
     logger_cls = logging.getLoggerClass()
     frame = inspect.currentframe()
     logger = None
-    try:
-        # frame may be None, see https://docs.python.org/2/library/inspect.html#inspect.currentframe
-        if frame is None:
-            framerecords = []
-        else:
-            framerecords = inspect.getouterframes(frame)
 
-        for frameinfo in framerecords:
-            bindings = inspect.getargvalues(frameinfo[0]).locals
-            for val in bindings.values():
-                if isinstance(val, logger_cls):
-                    logger = val
-    finally:
-        # make very sure that reference to frame object is removed, to avoid reference cycles
-        # see https://docs.python.org/2/library/inspect.html#the-interpreter-stack
-        del frame
+    # frame may be None, see https://docs.python.org/2/library/inspect.html#inspect.currentframe
+    if frame is not None:
+        try:
+            # consider calling stack in reverse order, i.e. most inner frame (closest to caller) first
+            for frameinfo in inspect.getouterframes(frame)[::-1]:
+                bindings = inspect.getargvalues(frameinfo[0]).locals
+                for val in bindings.values():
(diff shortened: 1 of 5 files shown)


lib/vsc/utils/exceptions.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    Get logger defined in caller's environment
+    @return: logger instance (or None if none was found)
+    """
+    logger_cls = logging.getLoggerClass()
+    frame = inspect.currentframe()
+    logger = None
+    try:
+        # frame may be None, see https://docs.python.org/2/library/inspect.html#inspect.currentframe
+        framerecords = (frame is None and []) or inspect.getouterframes(frame)
```
stdweird: 4 lines please

boegel:
removed the Py2.4 tests

boegel:
Jenkins: test this please

boegel:
@stdweird: remark fixed, was that the only one?

lib/vsc/utils/exceptions.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    Get logger defined in caller's environment
+    @return: logger instance (or None if none was found)
+    """
+    logger_cls = logging.getLoggerClass()
+    frame = inspect.currentframe()
+    logger = None
+    try:
+        # frame may be None, see https://docs.python.org/2/library/inspect.html#inspect.currentframe
+        if frame is None:
```
stdweird: the whole `try/finally` blck is part of the else (or rather the `if is not None`)

lib/vsc/utils/exceptions.py
```diff
@@ -0,0 +1,82 @@
...
+            framerecords = []
+        else:
+            framerecords = inspect.getouterframes(frame)
+
+        for frameinfo in framerecords:
+            bindings = inspect.getargvalues(frameinfo[0]).locals
+            for val in bindings.values():
+                if isinstance(val, logger_cls):
+                    logger = val
```
stdweird: add `break`?

lib/vsc/utils/exceptions.py
```diff
@@ -0,0 +1,82 @@
...
+        @param logger: logger to use
+        """
+        super(LoggedException, self).__init__(msg)
+
+        # try to use logger defined in caller's environment
+        if logger is None:
+            logger = get_callers_logger()
+            # search can fail, use root logger as a fallback
+            if logger is None:
+                logger = fancylogger.getLogger()
```
stdweird: make `fancylogger` a class constant. if someone want to override it, it's the easiest way

test/exceptions.py
```diff
@@ -0,0 +1,122 @@
...
+
+@author: Kenneth Hoste (Ghent University)
+"""
+import logging
+import os
+import re
+import tempfile
+from unittest import TestLoader, main
+
+from vsc.utils.exceptions import LoggedException
```
stdweird: test `get_callers_logger` itself if possible?

stdweird:
some minor remarks added

lib/vsc/utils/exceptions.py
```diff
@@ -0,0 +1,82 @@
...
+
+        logger.error(msg)
```
stdweird: also the log function should be configurable with class constant (probably needed in EB)

test/exceptions.py
boegel: done

boegel:
@stdweird: remarks fixed

lib/vsc/utils/exceptions.py
```diff
@@ -0,0 +1,86 @@
...
+    LOGGER_MODULE = fancylogger
+    # name of logging method to use (must accept a single argument of type string, i.e. the log message)
+    LOGGING_METHOD = 'error'
+
+    def __init__(self, msg, logger=None):
+        """
+        Constructor.
+        @param logger: logger to use
+        """
+        super(LoggedException, self).__init__(msg)
```
stdweird: is there anything in the `__init__` that could cause the logging part below not to be reached? (i guess only a `raise` in the `__init__` can cause that?)
to be safe, i'd move the `super` after the logging block (then there's always logging).
(maybe this is a bit farfetched :smile: )

stdweird:
@boegel one minor remark left.

lib/vsc/utils/exceptions.py
boegel: I doubt there's anything, but it can't hurt to do the logging first, so changed.

boegel:
@stdweird: remark fixed, pls recheck how `LOGGING_METHOD` is defined, I had to change that to redefine it in EasyBuild (to avoid an infinite (re)raise situation with `log.error`)

stdweird:
what do you do now in EB?

boegel:
@stdweird: see https://github.com/hpcugent/easybuild-framework/pull/1218/files#diff-1b40e5d23ba1913560f23697b64533cbR57

lib/vsc/utils/exceptions.py
```diff
@@ -0,0 +1,88 @@
...
+    """Exception that logs it's message when it is created."""
+
+    # logger module to use (must provide getLogger() function)
+    LOGGER_MODULE = fancylogger
+    # logging method to use
+    # takes a logger instance and an argument of type string, i.e. the log message
+    # note: 'self' (LoggedException instance) is passed as 1st argument, which is typically just ignored
+    LOGGING_METHOD = lambda _, logger, msg: logger.error(msg)
+
+    def __init__(self, msg, logger=None):
```
stdweird: support `*args`

lib/vsc/utils/exceptions.py
```diff
@@ -0,0 +1,88 @@
...
+
+class LoggedException(Exception):
+    """Exception that logs it's message when it is created."""
+
+    # logger module to use (must provide getLogger() function)
+    LOGGER_MODULE = fancylogger
+    # logging method to use
+    # takes a logger instance and an argument of type string, i.e. the log message
+    # note: 'self' (LoggedException instance) is passed as 1st argument, which is typically just ignored
+    LOGGING_METHOD = lambda _, logger, msg: logger.error(msg)
```
stdweird: why is this needed?

lib/vsc/utils/exceptions.py
```diff
@@ -0,0 +1,88 @@
...
+        @param logger: logger to use
+        """
+        # try to use logger defined in caller's environment
+        if logger is None:
+            logger = get_callers_logger()
+            # search can fail, use root logger as a fallback
+            if logger is None:
+                logger = self.LOGGER_MODULE.getLogger()
+
+        self.LOGGING_METHOD(logger, msg)
```
stdweird:
```python
getartr(logger, self.LOGGING_METHOD, logger.error)(msg, *args)
```

lib/vsc/utils/exceptions.py
boegel: before, this was just `'error'`, but that doesn't work in an EasyBuild context (since the `error` defined by EB raises `EasyBuildError`, which then results in a raise-log infinite loop)

lib/vsc/utils/exceptions.py
boegel: no it's not (anymore)

lib/vsc/utils/exceptions.py
boegel: ok, I'll tackle that

lib/vsc/utils/exceptions.py
boegel: see https://github.com/hpcugent/easybuild-framework/pull/1218/files#diff-1b40e5d23ba1913560f23697b64533cbR57, we need to pass a custom method in EasyBuild to avoid the infinite loop

stdweird: in EB, make a `_log_no_raise` method as part of `EasyBuildLog`, and set `LOGGING_METHOD = '_log_no_raise'`?

boegel: Hmm, ok, I'll see if that works out as expected/hoped. I *think* there was something that was blocking me from doing that, but I may be mixing things up.

boegel:
@stdweird: remarks fixed, should be good to go? clean approach for specifying logging method just based on name works out well in EasyBuild

stdweird:
@boegel ok, seems sufficient for EB so merging this in. nice work

PR #162, opened 2015-03-22
Add add_flex action and error logging for possible environment variable typos 

### LINKED ISSUE #161: feature request: support openpathlist for option type
`openpathlist` option type as an alternative to the existing `pathlist` option type, which allows prepending/appending rather than always overwriting

export EASYBUILD_ROBOT_PATHS=:/custom  # append to existing value
export EASYBUILD_ROBOT_PATHS=/custom:  # prepend to existing value
export EASYBUILD_ROBOT_PATHS=/custom  # overwrite existing value (like for 'pathlist'-typed option)

Would be useful in the context of EasyBuild to avoid overwriting `--robot-paths` default value.

cc @dominikborkowski

stdweird:
fixes #161

### DIFF
--- a/lib/vsc/utils/generaloption.py
+++ b/lib/vsc/utils/generaloption.py
@@ -106,6 +106,22 @@ def check_str_list_tuple(option, opt, value):
         return klass(split)
 
 
+def get_empty_add_flex(allvalues):
+    """Return the empty element for add_flex action for allvalues"""
+    empty = None
+    emptyisNone = False # in case None is the empty value
+
+    if isinstance(allvalues, (list, tuple)):
+        if isinstance(allvalues[0], basestring):
+            empty = ''
+
+    if (empty is None) and not emptyisNone:
+        msg = "get_empty_add_flex cannot determine empty element for type %s (%s)"
+        self.log.raiseException(msg % (type(allvalues), allvalues))
+
+    return empty
+
+
 class ExtOption(CompleterOption):
     """Extended options class
         - enable/disable support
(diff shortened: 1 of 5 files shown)


stdweird:
@boegel this should be sufficient for the EB issues you wanted

lib/vsc/utils/generaloption.py
```diff
@@ -106,6 +106,22 @@ def check_str_list_tuple(option, opt, value):
         return klass(split)
 
 
+def get_empty_add_flex(allvalues):
+    """Return the empty element for add_flex action for allvalues"""
+    empty = None
+    emptyisNone = False # in case None is the empty value
```
boegel: drop silly-named and not even used `emptisNone`

stdweird: it's there for future extensions

boegel: extensions of what?

don't include code you're not actually using, it looks silly

boegel:
still fixes and moar tests via https://github.com/stdweird/vsc-base/pull/4 (also fixing the merge conflict on the vsc-base version)

boegel:
@stdweird: I ran into a potential issue when writing up the unit tests for EB's `--robot-paths` that relies on this, see https://github.com/hpcugent/easybuild-framework/pull/1252/files#r27655849

stdweird:
@boegel rereview or merge

boegel:
ready to go in imho, but I'd like to see #165 and #166 merged first, to avoid more version bumps

stdweird:
@boegel bug is pretty stupid, not sure about the fix yet

boegel:
"Would you like to know more?" *clicks yes*

setup.py
```diff
@@ -52,7 +52,7 @@ def remove_bdist_rpm_source_file():
 
 PACKAGE = {
     'name': 'vsc-base',
-    'version': '2.1.2',
+    'version': '2.2.0',
```
boegel: this wil need to be bumped again, cfr #168

boegel:
fix for the bug that is causing this in https://github.com/stdweird/vsc-base/pull/6/files

boegel:
Going in (finally), thanks @stdweird for looking into this!

PR #164, opened 2015-04-02
Default mpi log format requires fancyrecord

stdweird:


lib/vsc/utils/fancylogger.py
```diff
@@ -341,15 +343,22 @@ def getLogger(name=None, fname=False, clsname=False, fancyrecord=None):
...
+        # it's easier to explain if we do it this way
+        fancyrecord = FANCYLOG_FANCYRECORD
+
     if name:
         nameparts.append(name)
-    elif fancyrecord is None or fancyrecord:  # only be fancy if fancyrecord is True or no name is given
+    elif fancyrecord is None:  # only be fancy if fancyrecord is True or no name is given
         fancyrecord = True
+
     fancyrecord = bool(fancyrecord)  # make sure fancyrecord is a nice bool, not None
```
boegel: drop this? seems to me fancyrecord can never be a non-bool anymore (unless something else than `None` or `True` or `False` is passed as a function argument)

test/fancylogger.py
```diff
@@ -369,6 +369,46 @@ def test_normal_logging(self):
         fancylogger.logToScreen(enable=False, handler=handler)
         sys.stderr = _stderr
 
+    def test_fancyrecord(self):
+        """
+        Test fancyrecord usage
+        """
+        logger=fancylogger.getLogger()
+        self.assertTrue(logger.fancyrecord == True)
```
boegel: use `assertEqual` when comparing values, which will lead to clear test failure messages

test/fancylogger.py
```diff
@@ -369,6 +369,46 @@ def test_normal_logging(self):
...
+
+        logger=fancylogger.getLogger('myname')
+        self.assertTrue(logger.fancyrecord == True)
+
+        logger=fancylogger.getLogger('myname', fancyrecord=False)
+        self.assertTrue(logger.fancyrecord == False)
+
+        fancylogger.FANCYLOG_FANCYRECORD=orig
```
boegel: ` = ` (spaces)

lib/vsc/utils/fancylogger.py
stdweird: so leave it there in case someone does?

test/fancylogger.py
```diff
@@ -369,6 +369,46 @@ def test_normal_logging(self):
         fancylogger.logToScreen(enable=False, handler=handler)
         sys.stderr = _stderr
 
+    def test_fancyrecord(self):
+        """
+        Test fancyrecord usage
+        """
+        logger=fancylogger.getLogger()
```
boegel: ` = ` (spaces)

test/fancylogger.py
```diff
@@ -369,6 +369,46 @@ def test_normal_logging(self):
...
+        logger=fancylogger.getLogger(fancyrecord=False)
+        self.assertTrue(logger.fancyrecord == False)
+
+        logger=fancylogger.getLogger('myname')
+        self.assertTrue(logger.fancyrecord == False)
+
+        logger=fancylogger.getLogger('myname', fancyrecord=True)
+        self.assertTrue(logger.fancyrecord == True)
+
+        orig=fancylogger.FANCYLOG_FANCYRECORD
```
boegel: ` = ` (spaces)

and better use this on top of the test method?

test/fancylogger.py
```diff
@@ -369,6 +369,46 @@ def test_normal_logging(self):
...
+        fancylogger.FANCYLOG_FANCYRECORD=False
```
boegel: ` = ` (spaces)

test/fancylogger.py
```diff
@@ -369,6 +369,46 @@ def test_normal_logging(self):
...
+
+        logger=fancylogger.getLogger()
```
boegel: ` = ` (spaces), same below

test/fancylogger.py
```diff
@@ -378,7 +418,6 @@ def tearDown(self):
         fancylogger.FancyLogger.RAISE_EXCEPTION_CLASS = self.orig_raise_exception_class
         fancylogger.FancyLogger.RAISE_EXCEPTION_LOG_METHOD = self.orig_raise_exception_method
 
-
```
boegel: please keep this?

boegel:
@stdweird: only minor remarks, thanks for tackling the tests too!

lib/vsc/utils/fancylogger.py
boegel: yeah, may be safest, I guess...

stdweird:
@boegel done

lib/vsc/utils/fancylogger.py
boegel: maybe add a test case that checks that things still work when passing a non-boolean value then?

```python
logger = fancylogger.getLogger('myname', fancyrecord='yes')
self.assertEqual(logger.fancyrecord, True)

logger = fancylogger.getLogger('myname', fancyrecord=0)
self.assertEqual(logger.fancyrecord, False)
```

boegel:
alrighty, going in!

PR #165, opened 2015-04-05
(optionally) include location info in LoggedException log message

boegel:
ported from https://github.com/hpcugent/easybuild-framework/pull/1249

### DIFF
--- a/lib/vsc/utils/exceptions.py
+++ b/lib/vsc/utils/exceptions.py
@@ -99,10 +99,7 @@ class LoggedException(Exception):
             # i.e. starting with an entry from LOC_INFO_TOP_PKG_NAMES
             path_parts = frameinfo[1].split(os.path.sep)
             top_indices = [path_parts.index(n) for n in self.LOC_INFO_TOP_PKG_NAMES if n in path_parts]
-            if top_indices:
-                relpath = os.path.join(*path_parts[max(top_indices):])
-            else:
-                relpath = os.path.join(*path_parts)
+            relpath = os.path.join(*path_parts[max(top_indices or [0]):])
 
             # include location info at the end of the message
             # for example: "Nope, giving up (at easybuild/tools/somemodule.py:123 in some_function)"
--- a/lib/vsc/utils/exceptions.py
+++ b/lib/vsc/utils/exceptions.py
@@ -98,9 +98,11 @@ class LoggedException(Exception):
             # determine short location of Python module where error was raised from,
             # i.e. starting with an entry from LOC_INFO_TOP_PKG_NAMES
             path_parts = frameinfo[1].split(os.path.sep)
-            relpath = path_parts.pop()
-            while path_parts and not any([relpath.startswith(n + os.path.sep) for n in self.LOC_INFO_TOP_PKG_NAMES]):
-                relpath = os.path.join(path_parts.pop() or os.path.sep, relpath)
+            top_indices = [path_parts.index(n) for n in self.LOC_INFO_TOP_PKG_NAMES if n in path_parts]
+            if top_indices:
+                relpath = os.path.join(path_parts[max(top_indices):])
+            else:
+                relpath = os.path.join(path_parts)
 
             # include location info at the end of the message
             # for example: "Nope, giving up (at easybuild/tools/somemodule.py:123 in some_function)"
--- a/lib/vsc/utils/exceptions.py
+++ b/lib/vsc/utils/exceptions.py
@@ -89,7 +89,7 @@ class LoggedException(Exception):
             frames_up = 1
             if self.__class__ != LoggedException:
                 # move a level up when this instance is derived from LoggedException
-                frames_up = 1
+                frames_up += 1
 
             # figure out where error was raised from
             # current frame: this constructor, one frame above: location where this EasyBuildError was created/raised
--- a/lib/vsc/utils/exceptions.py
+++ b/lib/vsc/utils/exceptions.py
@@ -31,6 +31,7 @@ Module providing custom exceptions.
 """
 import inspect
 import logging
+import os
 from vsc.utils import fancylogger
 
 
@@ -69,6 +70,8 @@ class LoggedException(Exception):
     # name of logging method to use
     # must accept an argument of type string, i.e. the log message, and an optional list of formatting arguments
     LOGGING_METHOD_NAME = 'error'
+    # list of top-level package names to use to format location info; None implies not to include location info
+    LOC_INFO_TOP_PKG_NAMES = []
 
     def __init__(self, msg, *args, **kwargs):
         """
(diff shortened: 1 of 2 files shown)


boegel:
@stdweird: please review

lib/vsc/utils/exceptions.py
```diff
@@ -81,6 +84,27 @@ def __init__(self, msg, *args, **kwargs):
...
 
+        if self.LOC_INFO_TOP_PKG_NAMES is not None:
+            # determine correct frame to fetch location information from
+            frames_up = 1
+            if self.__class__ != LoggedException:
+                # move a level up when this instance is derived from LoggedException
+                frames_up += 1
+
+            # figure out where error was raised from
+            # current frame: this constructor, one frame above: location where this EasyBuildError was created/raised
```
stdweird: `EasyBuildError` -> `LoggedException`

lib/vsc/utils/exceptions.py
```diff
@@ -81,6 +84,27 @@ def __init__(self, msg, *args, **kwargs):
...
+            frameinfo = inspect.getouterframes(inspect.currentframe())[frames_up]
+
+            # determine short location of Python module where error was raised from (starting with 'easybuild/' or 'vsc/')
```
stdweird: remove `easybuild / vsc`; explain the usage of `LOC_INFO_TOP_PKG_NAMES` instead

lib/vsc/utils/exceptions.py
```diff
@@ -81,6 +84,27 @@ def __init__(self, msg, *args, **kwargs):
...
+            path_parts = frameinfo[1].split(os.path.sep)
+            relpath = path_parts.pop()
+            while path_parts and not any([relpath.startswith(n + os.path.sep) for n in self.LOC_INFO_TOP_PKG_NAMES]):
```
stdweird: `relpath` can't have a `os.path.sep`, because it's an element of a `.split(os.path.sep)`, so the `any` is always false.
i have no idea what you are trying to do here tough

lib/vsc/utils/exceptions.py
```diff
@@ -81,6 +84,27 @@ def __init__(self, msg, *args, **kwargs):
...
+                relpath = os.path.join(path_parts.pop() or os.path.sep, relpath)
```
stdweird: what's the intention of `path_parts.pop() or os.path.sep`? can this ever occur at all?

lib/vsc/utils/exceptions.py
boegel: no, it's not always false, since `relpath` is updated via `os.path.join` in the body of the while loop

I'm using the `os.path.sep` as an 'end of name' marker to make sure I don't have a partial match, e.g.:

```python
'easybuild/tools/filetools.py'.startswith('easybuild/') == True
'easybuild-framework/easybuild/tools/filetools.py'.startswith('easybuild/') == False
```

What I'm doing here is trimming something like `/home/kehoste/easybuild-framework/easybuild/tools/filetools.py` down to just `easybuild/tools/filetools.py`

I agree this looks rather strange, but I haven't found a better to do this.

lib/vsc/utils/exceptions.py
boegel: `path_parts` is a list like `['', 'home', 'kehoste', 'easybuild-framework', 'easybuild', 'tools', 'filetools.py']`, so I'm taking the end of the list until I have something that starts with `easybuid/` (or `vsc/`)

boegel:
@stdweird: remarks fixed or replied to, please rereview

lib/vsc/utils/exceptions.py
```diff
@@ -81,6 +84,28 @@ def __init__(self, msg, *args, **kwargs):
...
+
+            # figure out where error was raised from
+            # current frame: this constructor, one frame above: location where this LoggedException was created/raised
+            frameinfo = inspect.getouterframes(inspect.currentframe())[frames_up]
+
+            # determine short location of Python module where error was raised from,
+            # i.e. starting with an entry from LOC_INFO_TOP_PKG_NAMES
+            path_parts = frameinfo[1].split(os.path.sep)
+            relpath = path_parts.pop()
+            while path_parts and not any([relpath.startswith(n + os.path.sep) for n in self.LOC_INFO_TOP_PKG_NAMES]):
```
stdweird: see previous remarks: relpath can't have a `os.path.sep`, so this is always false

lib/vsc/utils/exceptions.py
```diff
@@ -81,6 +84,28 @@ def __init__(self, msg, *args, **kwargs):
...
+                relpath = os.path.join(path_parts.pop() or os.path.sep, relpath)
```
stdweird: see previous remark: the `or` part can't be reached

lib/vsc/utils/exceptions.py
boegel: copy-pasting from reply to remark, which is hidden due to the updates:

no, it's not always false, since `relpath` is updated via `os.path.join` in the body of the while loop

I'm using the `os.path.sep` as an 'end of name' marker to make sure I don't have a partial match, e.g.:

```python
'easybuild/tools/filetools.py'.startswith('easybuild/') == True
'easybuild-framework/easybuild/tools/filetools.py'.startswith('easybuild/') == False
```

What I'm doing here is trimming something like `/home/kehoste/easybuild-framework/easybuild/tools/filetools.py` down to just `easybuild/tools/filetools.py`

I agree this looks rather strange, but I haven't found a better to do this.

`path_parts` is a list like `['', 'home', 'kehoste', 'easybuild-framework', 'easybuild', 'tools', 'filetools.py']`, so I'm taking the end of the list until I have something that starts with `easybuid/` (or `vsc/`)

stdweird: so it's like
```
path_parts = frameinfo[1].split(os.path.sep)
shortest = max([path_parts.index(n) for n in self.LOC_INFO_TOP_PKG_NAMES if n in path_parts])
relpath = os.path.join(path_parts[max:])
```
(and deal with empty list in `max` etc etc)

boegel: that works too, I like it

boegel:
@stdweird: remark fixed, please rereview

lib/vsc/utils/exceptions.py
```diff
@@ -81,6 +84,30 @@ def __init__(self, msg, *args, **kwargs):
...
+
+            # figure out where error was raised from
+            # current frame: this constructor, one frame above: location where this LoggedException was created/raised
+            frameinfo = inspect.getouterframes(inspect.currentframe())[frames_up]
+
+            # determine short location of Python module where error was raised from,
+            # i.e. starting with an entry from LOC_INFO_TOP_PKG_NAMES
+            path_parts = frameinfo[1].split(os.path.sep)
+            top_indices = [path_parts.index(n) for n in self.LOC_INFO_TOP_PKG_NAMES if n in path_parts]
+            if top_indices:
```
stdweird: replace if/else with
```python
relpath = os.path.join(*path_parts[max(top_indices or [0]):])
```
(or `+` instead of `or`; it doesn't really matter)

PR #166, opened 2015-04-06
add support for strlistdict option type (WIP)

boegel:
works, but could benefit from some cleanup (too much copy-pasting going on now)

when generating string representation of dict, keys must be iterated over sorted (to ensure unique output)

also, unclear why `--ext-strdict` isn't included in output of `generate_cmd_line` (see tests)

### DIFF
--- a/lib/vsc/utils/generaloption.py
+++ b/lib/vsc/utils/generaloption.py
@@ -90,6 +90,7 @@ def what_str_list_tuple(name):
 
     return sep, klass, helpsep
 
+
 def check_str_list_tuple(option, opt, value):
     """
     check function for strlist and strtuple type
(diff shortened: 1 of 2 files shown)


lib/vsc/utils/generaloption.py
```diff
@@ -106,6 +107,25 @@ def check_str_list_tuple(option, opt, value):
         return klass(split)
 
 
+def check_dict(option, opt, value):
+    """
+    Check function for dict type
```
stdweird: can we have a valid use case for such a horrible self invented issue? why not support JSON, at least it's well defined.

boegel: see https://github.com/hpcugent/easybuild-framework/pull/1257/files#r27827988

`eb --module-properties=lmod:sticky;arch:mic,offload ...`

boegel:
probably not going anywhere, so closing this (I'll leave the branch around)

PR #169, opened 2015-04-09
bump version to 2.1.3

boegel:
version bump is required because #165 was merged, and hpcugent/easybuild-framework#1249 requires a newer vsc-base version


boegel:
I'm merging this, it's only a version bump, and I have good reasons for it, so no harm done.

PR #170, opened 2015-04-09
improve fancylogger performance significantly via caching inspect.getsourcefile and getRootLogger

boegel:
Opening this PR as is so we don't forget about it, may need some further tweaking to make the caching optional for example.

This yields quite significant performance improvements.

To give some idea: the full EasyBuild framework test suite runs in ~780s with this in place, while it requires 1150s without (on my laptop on an SSD).

Using `python -m cProfile`, I can see:

* a drop in the number of `posix.stat` calls from 8.5M to 218K, i.e. ~40x less (the number of `posix.lstat` calls remained stable), thanks to the monkey-patched `inspect.getsourcefile`
* a drop in the number of `inspect.ismodule` calls from a whopping 410M to 900K, i.e. ~450x less, thanks to the caching of `getRootLogger()`

With the patch applied, most time is now spent on ~15K `select.select` calls (which is basically I/O wait stuff, see https://docs.python.org/2/library/select.html.

@stdweird, @JensTimmerman: please review

### DIFF
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -73,6 +73,19 @@ Logging to a udp server:
 """
 
 import inspect
+
+# patch inspect.getsourcefile to cache results to avoid tons of stat syscalls
+_cachegetsourcefile = {}
+_orig_inspect_getsourcefile = inspect.getsourcefile
+def cached_getsourcefile(obj):
+    """Faster version of inspect.getsourcefile, by caching results."""
+    fn = inspect.getfile(obj)
+    if not fn in _cachegetsourcefile:
+        _cachegetsourcefile[fn] = _orig_inspect_getsourcefile(obj)
+    return _cachegetsourcefile[fn]
+
+inspect.getsourcefile = cached_getsourcefile
+
 import logging
 import logging.handlers
 import os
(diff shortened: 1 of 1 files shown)


stdweird:
this PR is a hack at best, esp the `getRootLogger` needs to be understood better.
the inspect cache should be off by default, and controllable via function (enable the bypass e.g. after initialisation of the logger stuff).

JensTimmerman:
I'm not sure about all the side effects this code will have, 
I think the easiest wat go get better performance is to add the `if __debug__:` guard to hot spots, and when performance matters and debug info is not needed, run your code with `python -O`

boegel:
Some interesting info w.r.t. `-O` and `__debug__`: http://stackoverflow.com/questions/4777113/what-does-python-optimization-o-or-pythonoptimize-do

boegel:
With https://github.com/hpcugent/easybuild-framework/pull/1282, the time required for the EasyBuild framework unit tests on top of current vsc-base `master` is about 12m on my laptop; with this patch in place, it drops further until just 4.5m...

I'll look into reworking this using `__debug__`, and check if we can run EB with `python -O` by default.

boegel:
I guess this is mostly rendered useless by #182...

JensTimmerman:
caching could possibly still make sense in some cases, but I kind of prefer the `__debug__` way of doing things, since it's guaranteed to be consistent, either do the call all the time, or do it never.
Not really possible to end up in a weird state.

Another option here that I would prefer over the current caching:
get the root logger name once, at import time, and save it, then just use that one all the time instead of spending your time seeiing if it's cached or not.
But I believe this was the initial behaviour, and could give confusing results.

boegel:
replaced by `__debug__` approach in #182, thanks @JensTimmerman!

PR #173, opened 2015-05-08
(try to) detect --skip-build, and do not cleanup the previous build

### LINKED ISSUE #172: Cannot build fedora rpm - setup.py cleans up build
The canonical way python rpms are made in Fedora is:
~~~~~
%build
%{__python2} setup.py build

%install
%{__python2} setup.py install --skip-build --root %{buildroot}
~~~~~
This breaks with vsc-base because the first thing the install line does is cleanup the build:
~~~~~
+ /usr/bin/python2 setup.py install --skip-build --root /export/home/orion/redhat/BUILDROOT/python-vsc-base-2.2.0-1.fc21.x86_64
cleanup build
~~~~~
I can workaround by dropping --skip-build, but it would be nice if this was supported.

stdweird:
Fixes #172

### DIFF
--- a/lib/vsc/install/shared_setup.py
+++ b/lib/vsc/install/shared_setup.py
@@ -268,7 +268,20 @@ def build_setup_cfg_for_bdist_rpm(target):
 def action_target(target, setupfn=setup, extra_sdist=[]):
     # EXTRA_SDIST_FILES.extend(extra_sdist)
 
-    cleanup()
+    do_cleanup = True
+    try:
+        # very primitive check for install --skip-build
+        # in that case, we don't mind "leftover build";
+        # it's probably intentional
+        install_ind=sys.argv.index('install')
+        build_skip=sys.argv.index('--skip-build')
+        if build_skip > install_ind:
+            do_cleanup = False
+    except ValueError:
+        pass
+
+    if do_cleanup:
+        cleanup()
 
     build_setup_cfg_for_bdist_rpm(target)
     x = parse_target(target)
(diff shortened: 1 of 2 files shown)


lib/vsc/install/shared_setup.py
```diff
@@ -268,7 +268,20 @@ def build_setup_cfg_for_bdist_rpm(target):
...
 
-    cleanup()
+    do_cleanup = True
+    try:
+        # very primitive check for install --skip-build
+        # in that case, we don't mind "leftover build";
+        # it's probably intentional
+        install_ind=sys.argv.index('install')
+        build_skip=sys.argv.index('--skip-build')
+        if build_skip > install_ind:
```
boegel: why does order matter at all?

boegel:
missing version bump

lib/vsc/install/shared_setup.py
stdweird: as the comment above says, it's a very primitive check to see if the `--skip-build` is an `install` option

lib/vsc/install/shared_setup.py
```diff
@@ -268,7 +268,20 @@ def build_setup_cfg_for_bdist_rpm(target):
...
     # EXTRA_SDIST_FILES.extend(extra_sdist)
 
-    cleanup()
+    do_cleanup = True
+    try:
+        # very primitive check for install --skip-build
+        # in that case, we don't mind "leftover build";
+        # it's probably intentional
+        install_ind=sys.argv.index('install')
+        build_skip=sys.argv.index('--skip-build')
```
boegel: style: ` = ` (spaces), otherwise good to go I guess (I can't think of a better way unless changing the `setup.py` significantly more)

PR #174, opened 2015-05-12
Empty body fix

itkovian:
Functions must not have an empty body?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,7 @@ shared_setup.SHARED_TARGET.update({
 
 PACKAGE = {
     'name': 'vsc-base',
-    'version': '2.2.1',
+    'version': '2.2.2',
     'author': [sdw, jt, ag, kh],
     'maintainer': [sdw, jt, ag, kh],
     'packages': ['vsc', 'vsc.utils', 'vsc.install'],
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -52,7 +52,7 @@ def remove_bdist_rpm_source_file():
 
 PACKAGE = {
     'name': 'vsc-base',
-    'version': '2.2.1',
+    'version': '2.2.2',
```
boegel: is this really worth a version bump?

the `pass` should be there, for sure, but this is just a style cleanup (functionality was not changed)

wdpypere: if it is an rpm, it requires a bump every time. Otherwise you can't predict what is on the systems.

boegel: Sure, but this makes NO difference in terms of functionality. So why would we roll out a new RPM specifically for this?

wdpypere: that is another question. but I don't want 2 rpms with the same version and different checksums.

boegel: so then make this 2.2.2dev? (which should never be rolled out as an RPM)

wdpypere: we have no way of controlling that today. Also it does not pose that big of a problem/issue.

itkovian: It does make a large difference. If the original function would have been called, your code would bork. Now it won't.

boegel:
```
[kehoste:~] $ cat t.py 
def f():
    """f"""

def g():
    """g"""
    print 'in g'

g()
f()
[kehoste:~] $ python t.py
in g
```

boegel: the `pass` is included implicitly, so this is nothing more than a style fix...

PR #177, opened 2015-06-04
I made Log to file a bit sturdier against logging to non existing dirs HPC-5077

JensTimmerman:


test/fancylogger.py
```diff
@@ -51,6 +52,24 @@ def classless_function():
...
+class FansyLoggerLogToFileTest(EnhancedTestCase):
+    """
+    Tests for fancylogger, specific for logToFile
+    These dont' fit in the FansyLoggerTest class because they don't require the setUp and tearDown
+    """
+
+    def test_logtofile(self):
+        """Test to see if logtofile doesn't fail when logging to a non existing file /directory"""
+        tempdir = tempfile.gettempdir()
+        non_dir = os.path.join(tempdir, 'verytempdir')
```
boegel: use mkdtemp instead?

lib/vsc/utils/fancylogger.py
```diff
@@ -450,18 +450,24 @@ def logToFile(filename, enable=True, filehandler=None, name=None, max_bytes=MAX_
...
-                           enable=enable,
-                           handler=filehandler,
-                           )
+    handleropts = {
+        'filename': filename,
+        'mode': 'a',
+        'maxBytes': max_bytes,
+        'backupCount': backup_count,
+    }
+    # logging to a file is going to create the file later on, so let's try to be helpfull and create the path if needed
```
boegel: typo: `helpful`

and, is this only an optimisation, or does it fix a bug?

test/fancylogger.py
JensTimmerman: the point is that the directory should not exist yet.

test/fancylogger.py
```diff
@@ -51,6 +52,24 @@ def classless_function():
     logger.warn("from classless_function")
 
 
+class FansyLoggerLogToFileTest(EnhancedTestCase):
```
JensTimmerman: fanCy

lib/vsc/utils/fancylogger.py
```diff
@@ -450,18 +450,24 @@ def logToFile(filename, enable=True, filehandler=None, name=None, max_bytes=MAX_
...
+    directory = os.path.dirname(filename)
+    if not os.path.exists(directory):
+        os.makedirs(directory)
```
stdweird: check the result and add some clean error handling?

test/fancylogger.py
boegel: well, you make sure it doesn't below anyway? the two lines above are better covered by `mkdtemp`

lib/vsc/utils/fancylogger.py
JensTimmerman: it fixes the case where the code calling this function doesn't have to create the directory every time to be sure.
But that doesn't mean this will never give an error, just one small thing that doesn't need to be repeated all over the place.

lib/vsc/utils/fancylogger.py
JensTimmerman: Not sure what to do on an error, raise a FancyLoggerException?

lib/vsc/utils/fancylogger.py
JensTimmerman: And so yes, this fixes a case in some downstream projects, where we then don't need to create a dir in an rpm.

lib/vsc/utils/fancylogger.py
stdweird: maybe, but this might raise some OSerror which could surpise people. a reraise with a clearer error message is probbaly enough

JensTimmerman: it was already raising these errors, when trying to open the file, this step just makes it less likely to have it raised, should I also catch the os error that get's thrown when trying to write to a file that can't be created?  I explicitly didn't intend to catch any errors, just add this as a convinience so I don't have to always create the dir before calling logToFile, but logging to a file can always fail, and I don't think we should catch each individual case. E.g. When the user doesn't have permission to write tot he file, or the disk is full ( or over quota) or the filesystem is mounted read only, etc..
Each of these errors would have a different exception raised, and need a different action from the user. I don't want to go around and catch them all, to just raise them with a clearer message? 
Then what happens when a new type of filesystem is introduced that can throw new errors,  I need to keep the list of things that can fail up to date here?

stdweird: ofcourse not catching all possible errors, i meant something like
```
try:
    os.makedirs(directory)
except Exception, e:
    raise type(e)("Cannot create logdirectory %s: %s" % (directory, e))
```
this simply reraises with better message imho (stacktraces aren't that nice and friendly imho). and if the original exception is catched higher up, it's certainly not that easier.

boegel: this should do: `except IOError, OSError as err:`

I don't think you can run into other types of exceptions?

JensTimmerman: @boegel of course you can, read my argument again, new Exceptions can be invented in the future.
@stdweird this will hide the original stack-trace, making it harder to pinpoint the problem.
`    raise sys.exc_info()[0], (sys.exc_info()[1],"Cannot create logdirectory %s: %s" % (directory, e), sys.exc_info()[2]` might be an option, but It's not very clean imho.
(see https://docs.python.org/2/reference/simple_stmts.html#the-raise-statement, https://docs.python.org/2/library/sys.html#sys.exc_info )

boegel: Ah, so we're going to reinvent how we handle exceptions again, great.

Letting `os.makedirs` generate new exceptions would mean breaking the API, so it's not going to happen anytime soon.

And with a clear error message (i.e., one that includes `err`), a traceback is pretty useless in this case?

Creating the directory failed, which is most likely caused by a problem that's out of the control of the running script, so why produce a traceback?

Anyway...

JensTimmerman: what? I'm not following your argument here? Is this meant to stdweird's remarks, or as a remark to my latest remark?

I agree we shouldn't handle exceptions here, just let them flow...

And certainly not generate new exceptions, I completely agree there.

a traceback is the best way to later on figure out where exactly things broke, it is the second most usefull thing you could have, after the exceptions type.
I agree, creating the directory failed by a problem that's out of control of fancylogger, so why should it get in between there and claim to know something more then it does, in the current implementation it's not trying to do any of that, so please feel free to tell me this is correct ;-)

The newly proposed `raise` to come forward to stdweirds remarks would also not produce a traceback, just preserve the original one, and make sure the raised exception is left mostly intact with only an extra (potentially helpful?) message.

JensTimmerman:
remarks addressed

lib/vsc/utils/fancylogger.py
```diff
@@ -450,18 +450,28 @@ def logToFile(filename, enable=True, filehandler=None, name=None, max_bytes=MAX_
...
+        'maxBytes': max_bytes,
+        'backupCount': backup_count,
+    }
+    # logging to a file is going to create the file later on, so let's try to be helpful and create the path if needed
+    directory = os.path.dirname(filename)
+    if not os.path.exists(directory):
+        try:
+            os.makedirs(directory)
+        except Exception as ex:
+            raise sys.exc_info()[0], "Cannot create logdirectory %s: %s" % (directory, ex), sys.exc_info()[2]
```
JensTimmerman: to give you an example of what this changes:

```

~ $ python
Python 2.7.10 (default, May 29 2015, 14:44:58)
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
>>> from vsc.utils import fancylogger
>>> fancylogger.logToFile('/bla/foo')
  File "/usr/local/lib/python2.7/site-packages/vsc_base-2.2.3-py2.7.egg/vsc/utils/fancylogger.py", line 473, in logToFile
    handler=filehandler,
  File "/usr/local/lib/python2.7/site-packages/vsc_base-2.2.3-py2.7.egg/vsc/utils/fancylogger.py", line 520, in _logToSomething
    handler = handlerclass(**handleropts)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/handlers.py", line 117, in __init__
    BaseRotatingHandler.__init__(self, filename, mode, encoding, delay)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/handlers.py", line 64, in __init__
    logging.FileHandler.__init__(self, filename, mode, encoding, delay)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/__init__.py", line 911, in __init__
    StreamHandler.__init__(self, self._open())
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/__init__.py", line 941, in _open
    stream = open(self.baseFilename, self.mode)
IOError: [Errno 2] No such file or directory: '/bla/foo'
>>> fancylogger.logToFile('/tmp/bla/foo')
  File "/usr/local/lib/python2.7/site-packages/vsc_base-2.2.3-py2.7.egg/vsc/utils/fancylogger.py", line 473, in logToFile
    handler=filehandler,
  File "/usr/local/lib/python2.7/site-packages/vsc_base-2.2.3-py2.7.egg/vsc/utils/fancylogger.py", line 520, in _logToSomething
    handler = handlerclass(**handleropts)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/handlers.py", line 117, in __init__
    BaseRotatingHandler.__init__(self, filename, mode, encoding, delay)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/handlers.py", line 64, in __init__
    logging.FileHandler.__init__(self, filename, mode, encoding, delay)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/__init__.py", line 911, in __init__
    StreamHandler.__init__(self, self._open())
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/__init__.py", line 941, in _open
    stream = open(self.baseFilename, self.mode)
IOError: [Errno 2] No such file or directory: '/tmp/bla/foo'
>>>
 
 
new situation
~ $ python
Python 2.7.10 (default, May 29 2015, 14:44:58)
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
>>> from vsc.utils import fancylogger
>>> fancylogger.logToFile('/bla/foo')
  File "/usr/local/lib/python2.7/site-packages/vsc_base-2.2.3-py2.7.egg/vsc/utils/fancylogger.py", line 463, in logToFile
    os.makedirs(directory)
  File "/usr/local/Cellar/python/2.7.10/Frameworks/Python.framework/Versions/2.7/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: Cannot create logdirectory /bla: [Errno 13] Permission denied: '/bla'
>>>
~ $ python
Python 2.7.10 (default, May 29 2015, 14:44:58)
[GCC 4.2.1 Compatible Apple LLVM 6.1.0 (clang-602.0.53)] on darwin
>>> from vsc.utils import fancylogger
>>> fancylogger.logToFile('/tmp/bla/foo')
<logging.handlers.RotatingFileHandler object at 0x1047bb310>
```

boegel: @JensTimmerman: that's certainly a lot better, but typically the `logToFile` will be called by something else, which we don't really care about? So the traceback doesn't seem very useful to me (although it's certainly more useful than the one we got before).

lib/vsc/utils/fancylogger.py
```diff
@@ -450,18 +450,28 @@ def logToFile(filename, enable=True, filehandler=None, name=None, max_bytes=MAX_
...
+
+    return _logToSomething(
+        logging.handlers.RotatingFileHandler,
+        handleropts,
+        loggeroption='logtofile_%s' % filename,
```
boegel: shouldn't we sanitise `filename` here, since it may contain `/` ?

lib/vsc/utils/fancylogger.py
boegel: ok, I'm sick of disucssing this

Please only do one call to `sys.exc_info()`, and let's get this in.

test/fancylogger.py
boegel: ping, please use `mkdtemp`

test/fancylogger.py
```diff
@@ -51,6 +52,24 @@ def classless_function():
     logger.warn("from classless_function")
 
 
+class FancyLoggerLogToFileTest(EnhancedTestCase):
+    """
+    Tests for fancylogger, specific for logToFile
+    These dont' fit in the FancyLoggerTest class because they don't require the setUp and tearDown
```
boegel: they don't *require* it, but do they affect this test at all?

If not, please include `test_logtofile` in `FancyLoggerTest` instead, since this is a style-break (we usually only have one class per test module).

lib/vsc/utils/fancylogger.py
JensTimmerman: I'm not generating the traceback here, just preserving the original one, the fact that this is seen by the user is because there's no try-except around my call to logToFile, but as you can see, it will raise an error and print this traceback wether or not this code is in place, so this code is an improvement, and should be merged.

lib/vsc/utils/fancylogger.py
JensTimmerman: This doesn't seem to give any issues, it can be any random string.https://docs.python.org/2/library/functions.html#setattr

boegel: OK, fine then.

lib/vsc/utils/fancylogger.py
boegel: Sure, but let's only call `sys.exc_info()` once.

test/fancylogger.py
JensTimmerman: yes it affects this test, the setup already sets up a logger.

boegel: ok, then fix the docstring to state that, rather than "do not require"

PR #178, opened 2015-06-08
fixed chowing correct traceback for exceptions raised in raiseException

JensTimmerman:


boegel:
Nice @JensTimmerman 
For future reference: https://docs.python.org/2/reference/simple_stmts.html#grammar-token-raise_stmt

boegel:
Merging in before #177, so we can use the version bump there to also cover this.

PR #182, opened 2015-08-12
don't do inspect things in optimized mode (-O)

JensTimmerman:


ehiggs:
:+1: This is super important!

ehiggs:
```
$ python -m timeit -c 'from vsc.utils.generaloption import GeneralOption; GeneralOption(go_args=[])'
10 loops, best of 3: 12 msec per loop
```

JensTimmerman:
with patch:
```
-bash-4.1$ python -m timeit -c 'from vsc.utils.generaloption import GeneralOption; GeneralOption(go_args=[])'
10 loops, best of 3: 47 msec per loop
-bash-4.1$ python -O -m timeit -c 'from vsc.utils.generaloption import GeneralOption; GeneralOption(go_args=[])'
10 loops, best of 3: 3 msec per loop
```

Issue #184, opened 2015-08-17
options with 'store' action consume other options as values


boegel:
see https://github.com/hpcugent/easybuild/issues/113

stdweird:
so the issue is that the option parser should reject values that are valid options? and throw what error in this case? (and how would you like to pass the value `-D` to the option `-b`?)

boegel:
If `-D` is not considered a valid value for `-b`, then the error `'missing argument'` needs to be thrown, since `-b` requires an argument (since it's a `store` action).

I'm not sure if it's possible to discriminate, but I guess `-b=-D` would be acceptable, since that's not ambiguous.

stdweird:
makes sense. the error could make this clearer like `missing argument: trying to pass valid option -D is not allowed (use -b=-D instead)`, but i don't know if i can distinguish between `-b -D` and `-b=-D`.

boegel:
I'm sure you meant to say "I'm not sure how yet to distuingish between `-b -D` and `-b=-D`, but I can figure out a dirty hack" ;-)

stdweird:
reading the code, the shortoptions seem to have `-bVALUE` instead of `-b=VALUE`. only the longopts have `--longopt=VALUE`. raising the error should be trivial however.

boegel:
Well, if we can mention "use `--only-blocks=-D` instead" when someone is trying to use `-b -D`, I'm all for it.

stdweird:
hmm, it's a tiny bit more complicated then i tought, since this is only an issue on the commandline.

stdweird:
if using the longopt only (even when the commandline was shortopt) to force the option-as-value is acceptable, that might make things a bit simpler.

boegel:
You mean, only support `--only-blocks=-D` (but not the equivalent of `-b=-D`)? Sure, works for me.

ehiggs:
How does this work for store_int given `-D -7`

boegel:
@ehiggs: nice one :-)
I guess we'll need to make (another) exception for `store_int` (or maybe none at all, and just let it complain that `-x` is not a valid integer)?

stdweird:
how is `-7` a valid option?

boegel:
Well, it can be both a valid option (if you're crazy enough to use `-7` as a short option for `--seven`), and a valid value (for a `store_int` option), I guess.

stdweird:
well, you'll have to set `-x-7` or `--longx=-7` then

[...]

PR #185, opened 2015-08-18
Generaloption: handle option-as-value 

stdweird:
Fixes https://github.com/hpcugent/vsc-base/issues/184
Requires #193

### DIFF
--- a/lib/vsc/utils/generaloption.py
+++ b/lib/vsc/utils/generaloption.py
@@ -751,7 +751,7 @@ class ExtOptionParser(OptionParser):
 
     def get_env_options(self):
         """Retrieve options from the environment: prefix_longopt.upper()"""
-        if not self.environment_arguments is None:
+        if self.environment_arguments is not None:
             # already done, let's not do this again
             return self.environment_arguments
 
--- a/lib/vsc/utils/generaloption.py
+++ b/lib/vsc/utils/generaloption.py
@@ -751,6 +751,10 @@ class ExtOptionParser(OptionParser):
 
     def get_env_options(self):
         """Retrieve options from the environment: prefix_longopt.upper()"""
+        if not self.environment_arguments is None:
+            # already done, let's not do this again
+            return self.environment_arguments
+
         self.environment_arguments = []
 
         if not self.process_env_options:
--- a/lib/vsc/utils/generaloption.py
+++ b/lib/vsc/utils/generaloption.py
@@ -243,7 +243,7 @@ class ExtOption(CompleterOption):
                 else:
                     force_fmt = "%s" % self._short_opts[0]
 
-                errmsg = "Value '%s' is also a valid option, use '%s%s' instead." % (value, force_fmt, value))
+                errmsg = "Value '%s' is also a valid option, use '%s%s' instead." % (value, force_fmt, value)
                 raise OptionValueError(
 
         return Option.process(self, opt, value, values, parser)
--- a/lib/vsc/utils/generaloption.py
+++ b/lib/vsc/utils/generaloption.py
@@ -223,7 +223,7 @@ class ExtOption(CompleterOption):
                 self.log.raiseException("_set_attrs: unknown store_or %s" % self.store_or, exception=ValueError)
 
     def process(self, opt, value, values, parser):
-        """Handle option-as-value issues befoe continuing as usual"""
+        """Handle option-as-value issues before actually processing option."""
 
         if hasattr(parser, 'orig_rargs') and (value in parser._long_opt or value in parser._short_opt):
             orig_rargs = getattr(parser, 'orig_rargs')
@@ -243,9 +243,8 @@ class ExtOption(CompleterOption):
                 else:
                     force_fmt = "%s" % self._short_opts[0]
 
-                raise OptionValueError("Value '%s' is also a valid option."
-                                       " Use '%s%s' to pass this unambigously." %
-                                       (value, force_fmt, value))
+                errmsg = "Value '%s' is also a valid option, use '%s%s' instead." % (value, force_fmt, value))
+                raise OptionValueError(
 
         return Option.process(self, opt, value, values, parser)
 
@@ -497,7 +496,7 @@ class ExtOptionParser(OptionParser):
 
         self.environment_arguments = None
         self.commandline_arguments = None
-        self.orig_rargs = None # copy of the original arguments
+        self.orig_rargs = None  # copy of the original arguments
 
     def parse_args(self, args=None, values=None):
         """Keep a copy of the original arguments for callback / option processing"""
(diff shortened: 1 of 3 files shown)


stdweird:
@boegel this seems to works, but no unittests were added yet (and the current ones fail due to double logger somewhere, matser is fine)

lib/vsc/utils/generaloption.py
```diff
@@ -470,6 +496,15 @@ def __init__(self, *args, **kwargs):
 
         self.environment_arguments = None
         self.commandline_arguments = None
+        self.orig_rargs = None  # copy of the original arguments
+
+    def parse_args(self, args=None, values=None):
+        """Keep a copy of the original arguments for callback / option processing"""
+        # seems like self._get_args returns a copy, but better safe then sorry here
+        self.orig_rargs = self._get_args(args)[:]
```
boegel: I think if you make this `super(..., self)._get_args(args)`, then you can also remove the premature `return` I added in  `self._get_args`

boegel: well, close, `super` doesn't work because `OptionParser` doesn't derive from `object`

all tests pass with https://github.com/stdweird/vsc-base/pull/9

stdweird:
@boegel should be fine now

boegel:
Since #193 is merged in here, that should be reviewed/merged first, and this one should then be re-evaluated.

boegel:
@stdweird: please sync with `master` to trigger an updated diff/retest?

test/generaloption.py
```diff
@@ -764,6 +773,106 @@ def raise_error(msg, *args):
                               go_args=['--level-level'], envvar_prefix='GENERALOPTIONTEST', error_env_options=True,
                               error_env_option_method=raise_error)
 
+    def _match_testoption1_sysexit(self, args, msg):
+        self.mock_stderr(True)  # reset
+        self.assertErrorRegex(SystemExit, '.*', TestOption1, go_args=args)
+        stderr = self.get_stderr()
+        self.assertTrue(msg in stderr)
```
boegel: add `self.mock_stderr(False)` here to disable the mocking (so we can actually see something if something is broken hard :)

maybe the `self.assertErrorRegex` during mocking isn't a good idea either, will we see the actual failed test since `stderr` is being mocked?

maybe this is a good place for `try`/`except`/`finally` instead of `self.assertErrorRegex`:

```python
def _match_testoption1_sysexit(self, args, msg):
    """Check whether parsing supplied arguments with TestOption1 results in SystemExit error being raised."""
    system_exit = False
    self.mock_stderr(True)  # reset
    # purposely not using self.assertErrorRegex, since we're mocking stderr...
    try:
        TestOption(go_args=args)
    except SytemExit:
        system_exit = True
    finally:
        self.mock_stderr(False)
    self.assertTrue(system_exit)
```

boegel: ah, yes, and add the `self.assertTrue(msg in stderr)` too

test/generaloption.py
```diff
@@ -764,6 +773,106 @@ def raise_error(msg, *args):
...
+        self.assertEqual(tp.ALLOW_TYPO_AS_VALUE, True,
+                         msg="default do allow value similar to option")
+
+        # fake commandline args
+        # this is purely to test the function, actual usage is in test_option_as_value
+        tp.commandline_arguments = ['--base', '-something', 'base']
+        def t(fail, tpl):
+            for idx, value in enumerate(tp.commandline_arguments):
+                msg = tpl % value
+                res = fn(value, index=idx)
```
boegel: little point to use `fn`, except to obfuscate things? it's only used here...

test/generaloption.py
```diff
@@ -764,6 +773,106 @@ def raise_error(msg, *args):
...
+                         msg="default do not allow value as option")
+        self.assertEqual(tp.ALLOW_DASH_AS_VALUE, False,
+                         msg="default do not allow value starting with -")
+        self.assertEqual(tp.ALLOW_TYPO_AS_VALUE, True,
+                         msg="default do allow value similar to option")
+
+        # fake commandline args
+        # this is purely to test the function, actual usage is in test_option_as_value
+        tp.commandline_arguments = ['--base', '-something', 'base']
+        def t(fail, tpl):
```
boegel: don't use single-letter (inner) function names (and don't use `ttt` either ;))

boegel: maybe use `check`?

test/generaloption.py
```diff
@@ -764,6 +773,106 @@ def raise_error(msg, *args):
...
+        # -s requires an argument
+        self._match_testoption1_sysexit(['-b', '-s'],
+                                        "-s option requires an argument")
+
+        # valid ways of specifying '-b' as a value to --store/-s
+        topt = TestOption1(go_args=['--store=-b'])
+        self.assertEqual(topt.options.store, '-b')
+        self.assertFalse(topt.options.base)  # remain False (default value)
+
+        topt = TestOption1(go_args=['-s-b'])  # equivalent to --store=-b
```
boegel: hmm, is it a good idea to allow this (by default)?

`-s-b` is awfully close `-sb`/`-s -b`, and in 99% of the cases you probably intend the latter?

requiring using the long option to pass `-b` to `-s` (`--store`) makes sense?

test/generaloption.py
stdweird: i don't understand. isn't the setUp / tearDown supposed to do this? also don't understand the reamkr wrt exceptions and stderr.

test/generaloption.py
boegel: OK, let me clarify a bit more (this is Friday night after all).

Since `-b` is an existing option and `-s` requires a value, I don't think `-s-b` should be interpreted as `--store=-b` by default.

I would even take it further. If `-sb` is specified, I wouldn't consume `b` as a value since `-b` is a known short option.

Currently, in EasyBuild, you have to be very careful with `-r` (`--robot`), e.g. `eb -drfk` can be expected to mean `eb --debug --robot --force --skip`, but actually means `eb --debug --robot=f --skip` right now, which can lead to surprises/confusion.

If you really want pass a single letter directory as value to `-r`, requiring `--robot=f` makes sense to me and complaining if `-rf` is used makes sense to me.

boegel: Hmm, but then the issue remains if you really want to pass a value that starts with `-`.

I guess the only option that we should allow (by default) is `--store=-b`, since then it's clear it's (most likely) not a typo (due to the `--` and `=-` combo).

test/generaloption.py
boegel: While you're mocking stderr, you won't get any output on stderr anymore.

The tearDown will disable the mocking, sure, but it's cleaner to do that in here (since you may be calling this method several times in a single test).

test/generaloption.py
boegel: I'm working on a PR to support also disallowing `-sb`, but too tired to figure it out now (apparently). It's in the works though.

stdweird: please make it against master, not my PR

boegel: Well, I can make a new PR that includes your changes and mine together, sure.

But I think the diff in https://github.com/stdweird/vsc-base/pull/10 is pretty clear...

Anyway, opened #199

stdweird:
@boegel not going to merge #199 since i have no clue what else is in there and i'm not going to review it

PR #186, opened 2015-09-15
Support zipped egg installations of vsc-base

ehiggs:
One issue we find when running jobs on shared file systems is that the startup of processes takes a long time, mostly due to python hitting the file system quite aggressively. @boegel would like to try this out on the EasyBuild test suite and I'd like to support zipped egg installations in hod.

It turns out that vsc-base was happy to be installed as a zipped egg with only a modification to `modules_in_pkg_path`. It's pretty hairy, so I imagine there will be some useful feedback.

### DIFF
--- a/lib/vsc/utils/missing.py
+++ b/lib/vsc/utils/missing.py
@@ -46,6 +46,7 @@ import shlex
 import subprocess
 import sys
 import time
+import zipfile
 
 from vsc.utils import fancylogger
 from vsc.utils.frozendict import FrozenDict
@@ -332,6 +333,46 @@ def get_subclasses(klass, include_base_class=False):
     return get_subclasses_dict(klass, include_base_class=include_base_class).keys()
 
 
+def _extract_module_names(filelist):
+    """
+    Given a list of files, strip off the suffix ('.py') and hide _prefix
+    files except for __init__.py
+    """
+    module_regexp = re.compile(r"^(?P<modname>[^_].*|__init__)\.py$")
+    return [res.group('modname') for res in map(module_regexp.match, filelist) if res]
+
+
+def _trim_leading_sep(path):
+    """
+    If the first char is a path separator (/), return the path with the
+    separator trimmed off.
+    """
+    if path.startswith(os.sep):
+        return path[1:]
+    return path
+
+def _find_zipped_modules(zip_path, pkg_path):
+    """
+    Given a path to a zipped egg, and a pkg path, return the submodules inside
+    the module
+    """
+    zf = zipfile.ZipFile(zip_path)
+    names = zf.namelist()
+    prefix = os.path.commonprefix([zip_path, pkg_path])
+    zip_pkg_path = _trim_leading_sep(pkg_path[len(prefix):])
+    names = [_trim_leading_sep(n[len(zip_pkg_path):]) for n in names if n.startswith(zip_pkg_path)]
+
+    if names:
+        if '__init__.py' not in names:
+            tup = (zip_path, zip_pkg_path)
+            _log.debug('No __init__.py found in %s/%s. It is not a valid path' % tup)
+            return None
+        modules = _extract_module_names(names)
+        _log.debug("List of modules for package in %s: %s" % (pkg_path, modules))
+        return modules
+    return None
+
+
 def modules_in_pkg_path(pkg_path):
     """Return list of module files in specified package path."""
     # if the specified (relative) package path doesn't exist, try and determine the absolute path via sys.path
@@ -355,7 +396,10 @@ def modules_in_pkg_path(pkg_path):
                 if is_pkg:
                     newpath = abspath
                     break
-
+            elif zipfile.is_zipfile(sys_path_dir):
+                modules = _find_zipped_modules(sys_path_dir, pkg_path)
+                if modules is not None:
+                    return modules
         if newpath is None:
             # give up if we couldn't find an absolute path for the imported package
             tup = (pkg_path, sys.path)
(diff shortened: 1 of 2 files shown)


boegel:
Jenkins: ok to test

setup.py
```diff
@@ -58,6 +58,7 @@ def remove_bdist_rpm_source_file():
     'packages': ['vsc', 'vsc.utils', 'vsc.install'],
     'scripts': ['bin/logdaemon.py', 'bin/startlogdaemon.sh', 'bin/bdist_rpm.sh', 'bin/optcomplete.bash'],
     'install_requires' : ['setuptools'],
+    'zip_safe': True,
```
stdweird: what happens with the files in the rpm when this is set? also zipped?

lib/vsc/utils/missing.py
```diff
@@ -332,6 +333,46 @@ def get_subclasses(klass, include_base_class=False):
...
+def _extract_module_names(filelist):
+    """
+    Given a list of files, strip off the suffix ('.py') and hide _prefix
+    files except for __init__.py
+    """
+    module_regexp = re.compile(r"^(?P<modname>[^_].*|__init__)\.py$")
+    return [res.group('modname') for res in map(module_regexp.match, filelist) if res]
+
+
+def _trim_leading_sep(path):
```
stdweird: like `path.lstrip(os.sep)`, but with a dedicated function?

lib/vsc/utils/missing.py
```diff
@@ -332,6 +333,46 @@ def get_subclasses(klass, include_base_class=False):
     return get_subclasses_dict(klass, include_base_class=include_base_class).keys()
 
 
+def _extract_module_names(filelist):
+    """
+    Given a list of files, strip off the suffix ('.py') and hide _prefix
+    files except for __init__.py
+    """
+    module_regexp = re.compile(r"^(?P<modname>[^_].*|__init__)\.py$")
```
stdweird: should be a module constant

lib/vsc/utils/missing.py
```diff
@@ -332,6 +333,46 @@ def get_subclasses(klass, include_base_class=False):
     return get_subclasses_dict(klass, include_base_class=include_base_class).keys()
 
 
+def _extract_module_names(filelist):
+    """
+    Given a list of files, strip off the suffix ('.py') and hide _prefix
```
stdweird: `files`->`filesnames` (i.e. not a path)

lib/vsc/utils/missing.py
```diff
@@ -332,6 +333,46 @@ def get_subclasses(klass, include_base_class=False):
...
+    return path
+
+def _find_zipped_modules(zip_path, pkg_path):
+    """
+    Given a path to a zipped egg, and a pkg path, return the submodules inside
+    the module
+    """
+    zf = zipfile.ZipFile(zip_path)
+    names = zf.namelist()
+    prefix = os.path.commonprefix([zip_path, pkg_path])
```
stdweird: you aware of the following:
```
>>> os.path.commonprefix(['/ab','/ac'])
'/a'
```
not sure you want this

lib/vsc/utils/missing.py
```diff
@@ -332,6 +333,46 @@ def get_subclasses(klass, include_base_class=False):
...
+    zip_pkg_path = _trim_leading_sep(pkg_path[len(prefix):])
+    names = [_trim_leading_sep(n[len(zip_pkg_path):]) for n in names if n.startswith(zip_pkg_path)]
+
+    if names:
```
stdweird: any example where this would be empty? like an empty egg

lib/vsc/utils/missing.py
```diff
@@ -332,6 +333,46 @@ def get_subclasses(klass, include_base_class=False):
...
+        if '__init__.py' not in names:
+            tup = (zip_path, zip_pkg_path)
+            _log.debug('No __init__.py found in %s/%s. It is not a valid path' % tup)
+            return None
+        modules = _extract_module_names(names)
```
stdweird: shouldn't this be tested for empty list? (far more likely to be emtpy than `names` imho)

lib/vsc/utils/missing.py
```diff
@@ -364,8 +408,15 @@ def modules_in_pkg_path(pkg_path):
             pkg_path = newpath
             _log.debug("Found absolute package path %s" % pkg_path)
 
-    module_regexp = re.compile(r"^(?P<modname>[^_].*|__init__)\.py$")
-    modules = [res.group('modname') for res in map(module_regexp.match, os.listdir(pkg_path)) if res]
+    # Maybe it's a zipped egg.
+    else:
+        for sys_path_dir in sys.path:
+            _log.debug('Zipped egg found in system path: "%s". Checking if the the module there.' % sys_path_dir)
```
stdweird: this log statement should be after the `if`

stdweird:
@ehiggs unittests please

stdweird:
@ehiggs also, what's the case where `startup of processes takes a long time, mostly due to python hitting the file system quite aggressively`?

lib/vsc/utils/missing.py
ehiggs: isn't `_extract_module_names` just stripping off the `.py` and prefix `_`? So if we pass in names that we're confident exist and are part of a module (since there's an `__init__.py`) file then modules shouldn't be empty (since there's at least `__init__.py`).

lib/vsc/utils/missing.py
ehiggs: An egg where the module isn't found. e.g. looking for `easybuild/easyconfigs` in `/path/to/pkgs/hod.egg` In this case, `hod.egg` will have a common prefix, but the filter clause of the list comprehension will show that no modules in the egg start with `easybuild/easyconfigs`.

lib/vsc/utils/missing.py
ehiggs: Agreed.

```python
>>> os.path.dirname(os.path.commonprefix(['/ab', '/ac']))
'/'
>>> os.path.dirname(os.path.commonprefix(['/a/b', '/a/c']))
'/a'
```

boegel:
@ehiggs: please clarify why you close this

ehiggs:
#187 was opened to replace it.

PR #188, opened 2015-09-19
support using 'inspect' when desired in LoggedException

boegel:
I need this for https://github.com/hpcugent/easybuild-framework/pull/1357, since I want to make sure that an `EasyBuildException` that gets raised always includes the location where it was raised from.

See also https://github.com/hpcugent/easybuild-framework/pull/1357/files#r39891748

### DIFF
--- a/lib/vsc/utils/exceptions.py
+++ b/lib/vsc/utils/exceptions.py
@@ -82,7 +82,10 @@ class LoggedException(Exception):
         @param msg: exception message
         @param *args: list of formatting arguments for exception message
         @param logger: logger to use
+        @param include_location: include location where error was raised from
         """
+        include_location = kwargs.get('include_location', False)
+
         # format message with (optional) list of formatting arguments
         if args:
             msg = msg % args
@@ -94,7 +97,7 @@ class LoggedException(Exception):
                 # move a level up when this instance is derived from LoggedException
                 frames_up += 1
 
-            if __debug__:
+            if include_location or __debug__:
                 # figure out where error was raised from
                 # current frame: this constructor, one frame above: location where LoggedException was created/raised
                 frameinfo = inspect.getouterframes(inspect.currentframe())[frames_up]
--- a/lib/vsc/utils/exceptions.py
+++ b/lib/vsc/utils/exceptions.py
@@ -35,13 +35,13 @@ import os
 from vsc.utils import fancylogger
 
 
-def get_callers_logger(use_inspect=False):
+def get_callers_logger():
     """
     Get logger defined in caller's environment
     @return: logger instance (or None if none was found)
     """
     logger_cls = logging.getLoggerClass()
-    if use_inspect or __debug__:
+    if __debug__:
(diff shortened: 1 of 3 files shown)


boegel:
Jenkins: test this please

lib/vsc/utils/exceptions.py
```diff
@@ -82,7 +82,10 @@ def __init__(self, msg, *args, **kwargs):
         @param msg: exception message
         @param *args: list of formatting arguments for exception message
         @param logger: logger to use
+        @param include_location: include location where error was raised from
         """
+        include_location = kwargs.get('include_location', False)
```
stdweird: if you always want to set it for a certain subclass, why not use a class constant? is there a use case where you would to set if for individual exceptions?

boegel:
Jenkins: test this please

lib/vsc/utils/exceptions.py
boegel: class constant works for me, will change

lib/vsc/utils/exceptions.py
```diff
@@ -35,13 +35,13 @@
 from vsc.utils import fancylogger
 
 
-def get_callers_logger():
+def get_callers_logger(use_inspect=False):
```
stdweird: how does this work? ( i mean how is `use_inspect` set/passed/...)

boegel: `get_callers_logger(use_inspect=True)`, how else?

it's not set in the call done from `LoggedException`, which is fine, but having control is nice?

lib/vsc/utils/exceptions.py
boegel: now using class constant

lib/vsc/utils/exceptions.py
stdweird: yes, but adding undocumented code that noone uses aint.

lib/vsc/utils/exceptions.py
```diff
@@ -75,6 +75,8 @@ class LoggedException(Exception):
     LOGGING_METHOD_NAME = 'error'
     # list of top-level package names to use to format location info; None implies not to include location info
     LOC_INFO_TOP_PKG_NAMES = []
+    # include location where error was raised from
+    INCLUDE_LOCATION = False
```
stdweird: shouldn't the default be `__debug__`?

lib/vsc/utils/exceptions.py
boegel: ok, fine, I'll kick it out

lib/vsc/utils/exceptions.py
boegel: that works too, will change

boegel:
@stdweird: remarks fixed

JensTimmerman:
wow, this is not the right way to go, either you're running in debug mode, and you then you accept the performance hit, or you're not running in debug mode, and then the location is of no interest to you.

PR #189, opened 2015-09-21
Revert "support using 'inspect' when desired in LoggedException"

JensTimmerman:
Reverts hpcugent/vsc-base#188

This is the wrong way to go.


JensTimmerman:
well, it's only used in the logged-exception which you probably don't want to use anyway.

PR #191, opened 2015-09-21
enable zip_safe, and fix broken tests

boegel:
With some minor changes in the tests, vsc-base is zip-safe.

old description:
```
The biggest change here is moving the tests to the `vsc.test.vsc_base` package, such that:

i) the tests become part of the installation, which sort of required to run the tests on an installed egg
ii) clashes with other `test` namespaces (like the one in the Python stdlib) are avoided

So, Jenkins will have to be updated accordingly, to run the tests with `python -m vsc.test.vsc_base.suite`, *after* installing `vsc-base` as a zipped eggs in a temporary location.
Of course, the tests should still work when being run as `python lib/vsc/test/vsc_base/suite.py` too, so Jenkins should check both?

Overview of tests that ~~will need~~ needed to be fixed (probably because they rely on using `__file__` somehow) is available at ~~#190~~.~~
```


setup.py
```diff
@@ -55,8 +56,9 @@ def remove_bdist_rpm_source_file():
     'version': '2.2.6',
     'author': [sdw, jt, ag, kh],
     'maintainer': [sdw, jt, ag, kh],
-    'packages': ['vsc', 'vsc.utils', 'vsc.install'],
+    'packages': ['vsc', 'vsc.install', 'vsc.test', 'vsc.utils'],
```
boegel: maybe I should move the tests one level deeper even, to `vsc.test.vsc_base`, to avoid clashes with the other `vsc` packages when they move to this scheme as well?

boegel:
All tests work now, both with `python -m vsc.test.vsc_base.suite` (egg or non-egg) and with `python lib/vsc/test/vsc_base/suite.py`.

No code in `lib/vsc` was touched.

boegel:
@stdweird, @JensTimmerman: please review?

stdweird:
we should not ship unittests as part of any package. promoting the unittests to part of a namespace to work around some issue is not a solution, it's a pretty dirty hack at best.

setup.py
```diff
@@ -46,17 +48,19 @@ def remove_bdist_rpm_source_file():
...
 
 
 PACKAGE = {
     'name': 'vsc-base',
-    'version': '2.2.6',
+    'version': '2.3.0',
     'author': [sdw, jt, ag, kh],
     'maintainer': [sdw, jt, ag, kh],
-    'packages': ['vsc', 'vsc.utils', 'vsc.install'],
+    'packages': ['vsc', 'vsc.install', 'vsc.test', 'vsc.test.vsc_base', 'vsc.utils'],
```
stdweird: no.

boegel:
I don't see why shipping tests is wrong.
In fact, I think it's better, since it allows people to easily run the tests on their system. We ship the tests for the EasyBuild framework with EB itself too (only framework though, but that's mostly because of the namespace issue with using a top-level `test` namespace, so I'd like to change it there too).

I guess including the tests *in* the `vsc-base` `.egg` is not strictly needed though. They can simply be run as `python test/suite.py`, and 'talk' to a vsc-base that is installed as an egg. Making them work in both types of installations is far more important.

I'm happy to move them back again, but it'll make things a bit more tedious than they need to be (imho).

boegel:
@stdweird: moved back the tests, running them both with zipped eggs and 'normal' install still works

but we should update the Jenkins PR builder to also checked the zipped installation before merging this

stdweird:
the main reason not to ship them is that they are not part of the code. if convenience is so mportant, why don't you also ship all possible python binaries and all their dependencies with EB, will also make things easier.
if you are debugging something, you should accept the fact that you might need to install extra things simply to debug the specific issue, and once it's fixed, you shouldn't need it anymore.
i think that by shipping the unittests, and running them as part of EB, you are using the EB unittests as a way to verify the system EB runs on, rather then unittests of EB itself.

test/exceptions.py
```diff
@@ -170,7 +170,7 @@ def raise_testexception(msg, *args, **kwargs):
         self.assertErrorRegex(LoggedException, 'BOOM', raise_testexception, 'BOOM')
         logToFile(tmplog, enable=False)
 
-        log_re = re.compile(r"^%s :: BOOM \(at /.*/vsc/utils/testing.py:[0-9]+ in assertErrorRegex\)$" % rootlogname)
```
stdweird: out of curiosity, why was this needed? was the path cut off or something?

stdweird:
LGTM, i'll let @JensTimmerman merge it in

test/exceptions.py
boegel: yes, when in an egg, the printed path is a relative directory like `build/.../egg/...`

this is the error I was getting:

```
======================================================================
FAIL: test_loggedexception_location (test.exceptions.ExceptionsTest)
Test inclusion of location information in log message for LoggedException.
  File "build/bdist.macosx-10.10-intel/egg/vsc/test/exceptions.py", line 175, in test_loggedexception_location
    self.assertTrue(log_re.match(logtxt), "%s matches %s" % (log_re.pattern, logtxt))
AssertionError: ^runpy :: BOOM \(at /.*/vsc/utils/testing.py:[0-9]+ in assertErrorRegex\)$ matches runpy :: BOOM (at build/bdist.macosx-10.10-intel/egg/vsc/utils/testing.py:77 in assertErrorRegex)
```

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
Jenkins: test this please

boegel:
Jenkins setup has been changed to also check the tests for zipped-eggs installs

boegel:
@JensTimmerman: please review/merge?

setup.py
```diff
@@ -46,16 +47,17 @@ def remove_bdist_rpm_source_file():
...
-    'download_url': 'https://github.com/hpcugent/vsc-base'
+    'download_url': 'https://github.com/hpcugent/vsc-base',
+    'zip_safe': True,
 })
 
 
 PACKAGE = {
     'name': 'vsc-base',
-    'version': '2.2.6',
+    'version': '2.3.0',
```
JensTimmerman: no need to bump the version to 2.3 , nothing in the code changed. You can just bump the release, since that's the only thing that changed.

test/runner.py
```diff
@@ -2,8 +2,6 @@
 import os
 import sys
 
-sys.path.insert(0, os.path.dirname(os.path.dirname(__file__)))
```
JensTimmerman: this helps a lot in not accidentaly testing the installed version instead of the version in your current directory.

boegel: are you asking me to keep this around, or giving a :+1:?

there is no installed version of the tests...

setup.py
boegel: are you asking me to introduce a `'release': '1'`?

test/runner.py
JensTimmerman: no, I mean, if you have installed an old version of vsc-base but developing on a new version, this will make sure you're not accedentaly testing the installed version. But I guess that's exactly what you don't want to happen on jenkins?

setup.py
JensTimmerman: no, that doesn't work in rpm's, 2.2.7 is what I mean. 
It's fine for now, until we have an official version numbering policy I guess.

test/runner.py
boegel: This line was there to add the directory in which the `test` directory is located to the Python search path (so `python -m test.runner` works), this has nothing to do with the `vsc` namespace at all (which is in `lib/vsc`), so it only concerns the tests.

The vsc-base installation you're testing is controlled via `$PYTHONPATH`

PR #193, opened 2015-10-08
Support unittesting via 'python setup.py test'

### LINKED ISSUE #192: Add `python setup.py test` lib shim custom command.
See https://github.ugent.be/hpcugent/clusterbuildrpm-server/pull/18

stdweird:
fixes #192


stdweird:
retest this please

stdweird:
last tests ran with adapted jenkins to use `setup.py test`

setup.py
```diff
@@ -32,35 +32,35 @@
...
-shared_setup.SHARED_TARGET.update({
-    'url': 'https://github.com/hpcugent/vsc-base',
-    'download_url': 'https://github.com/hpcugent/vsc-base',
-    'zip_safe': True,
-})
 
 
 PACKAGE = {
     'name': 'vsc-base',
-    'version': '2.3.0',
```
wdpypere: no more version?

setup.py
```diff
@@ -32,35 +32,35 @@
...
     'author': [sdw, jt, ag, kh],
     'maintainer': [sdw, jt, ag, kh],
     'packages': ['vsc', 'vsc.install', 'vsc.utils'],
     'scripts': ['bin/logdaemon.py', 'bin/startlogdaemon.sh', 'bin/bdist_rpm.sh', 'bin/optcomplete.bash'],
     'install_requires' : ['setuptools'],
+    'zip_safe': True,
```
itkovian: Are we sure about this? We have vsc-utils that also install in the same namespace, or am I mixing things up?

setup.py
boegel: @wdpypere: psst, there's a line below this one ;)

setup.py
stdweird: i didn't add this...

lib/vsc/install/shared_setup.py
```diff
@@ -20,23 +20,46 @@
...
+# available authors
+ag = ('Andy Georges', 'andy.georges@ugent.be')
+jt = ('Jens Timmermans', 'jens.timmermans@ugent.be')
+kh = ('Kenneth Hoste', 'kenneth.hoste@ugent.be')
+lm = ('Luis Fernando Munoz Meji?as', 'luis.munoz@ugent.be')
+sdw = ('Stijn De Weirdt', 'stijn.deweirdt@ugent.be')
+wdp = ('Wouter Depypere', 'wouter.depypere@ugent.be')
+kw = ('Kenneth Waegeman', 'Kenneth.Waegeman@UGent.be')
+eh = ('Ewan Higgs', 'Ewan.Higgs@UGent.be')
+wp = ('Ward Poelmans', 'Ward.Poelmans@UGent.be')
```
itkovian: Maybe keep them ordered alphabetically?

setup.py
stdweird: nice catch, restored and bumped

setup.py
itkovian: According to the diff you did :)

stdweird: look a few lines back

boegel: @ageorges: shouldn't matter whether vsc-utils touches `vsc/utils` too, vsc-base by itself works just fine when installed as a zipped egg

However, other stuff that requires it as a dep doesn't support using it when it's installed as a zipped egg yet *cough* EasyBuild *cough*

So, for now, Jenkins shouldn't install it as a zipped egg after a successful merge/test, i.e. it should use `easy_install -U`

test/runner.py
```diff
@@ -2,6 +2,9 @@
 import os
 import sys
 
+if __name__ == '__main__':
+    sys.path.insert(0, os.path.dirname(os.path.dirname(__file__)))
```
boegel: this was kicked out in a previous PR, why add it again?

if it's really needed (it shouldn't), please add a clear comment here

lib/vsc/install/shared_setup.py
```diff
@@ -20,23 +20,46 @@
...
+
+# FIXME: do we need this here? it won;t hurt, but still ...
+REGEXP_REMOVE_SUFFIX = re.compile(r'(\.(?:py|sh|pl))$')
```
ehiggs: If we don't mind splitting any ends, we can use `'some-string'.rsplit('.', 1)[0]`

test/runner.py
stdweird: it will break again if we remove the hack from setup.py

lib/vsc/install/shared_setup.py
```diff
@@ -166,13 +175,56 @@ def find_package_modules (self, package, package_dir):
...
         orig_bdist_rpm.run(self)
 
+class VscTestCommand(TestCommand):
+    """
+    The cmdclass for testing
+    """
+    def run_tests(self):
+        # should be setup.py
+        setup_py = inspect.stack()[-1][1]
+        log.info('run_tests from %s' % setup_py)
```
JensTimmerman: do we really need inspect here? it is kind of slow, and does not work if the .py file is not available (like when you only have the .pyc or pyo files)
why not use `sys.argv[0]`? Seems to do the same thing to me in these cases:

```
[jens@exile temp]$ vim bla.py
[jens@exile temp]$ python foo.py 
foo.py
foo.py


[jens@exile temp]$ python bla.py
bla.py
bla.py


[jens@exile temp]$ cat foo.py 
import bla
[jens@exile temp]$ cat bla.py
import inspect
import sys
import os
print inspect.stack()[-1][1]
print sys.argv[0] 
print os.path.dirname(sys.argv[0])
print os.path.dirname(inspect.stack()[-1][1])
```

stdweird: i pulled the inspect code form stackoverflow, rereading it, they tried to solve something more complicated, so i'll revert to `sys.argv[0]`

stdweird:
@boegel @JensTimmerman i added support to filter tests

lib/vsc/install/shared_setup.py
```diff
@@ -24,19 +24,46 @@
 import shutil
 import sys
 import re
+
+try:
+    from unittest.suite import TestSuite
+except ImportError:
+    from unittest import TestSuite
```
boegel: the fallback is for older Python versions or something? add a comment to clarify please

lib/vsc/install/shared_setup.py
```diff
@@ -24,19 +24,46 @@
...
+# available authors
+ag = ('Andy Georges', 'andy.georges@ugent.be')
+eh = ('Ewan Higgs', 'Ewan.Higgs@UGent.be')
+jt = ('Jens Timmermans', 'jens.timmermans@ugent.be')
+kh = ('Kenneth Hoste', 'kenneth.hoste@ugent.be')
+kw = ('Kenneth Waegeman', 'Kenneth.Waegeman@UGent.be')
+lm = ('Luis Fernando Munoz Meji?as', 'luis.munoz@ugent.be')
+sdw = ('Stijn De Weirdt', 'stijn.deweirdt@ugent.be')
+wdp = ('Wouter Depypere', 'wouter.depypere@ugent.be')
+wp = ('Ward Poelmans', 'Ward.Poelmans@UGent.be')
```
boegel: Hmm, are you using a different alphabet than the rest of us? ;)

lib/vsc/install/shared_setup.py
```diff
@@ -24,19 +24,46 @@
...
+
+# FIXME: do we need this here? it won;t hurt, but still ...
+REGEXP_REMOVE_SUFFIX = re.compile(r'(\.(?:py|sh|pl))$')
```
boegel: do I dare ask about this?

test/runner.py
boegel: This hack is not (really) related to the one in `setup.py`, I think...

This one is for the `import test`s below, the one in `setup.py` is for `import vsc.install` (since the latter insert `.../lib`)

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+    })
+
+
+def filter_testsuites(testsuites):
+    """(Recursive) filtering of (suites of) tests"""
+    test_filter = getattr(__builtin__,'__test_filter')['function']
+
+    res = type(testsuites)()
+
+    for t in testsuites:
```
boegel: don't use single-letter variables outside of list comprehensions pleae

lib/vsc/install/shared_setup.py
```diff
@@ -24,19 +24,46 @@
...
 from distutils import log  # also for setuptools
 from distutils.dir_util import remove_tree
 
+from setuptools.command.test import test as TestCommand
+from setuptools.command.test import ScanningLoader
+
 # 0 : WARN (default), 1 : INFO, 2 : DEBUG
 log.set_verbosity(2)
 
 has_setuptools = None
```
boegel: well, `setuptools` is there since we hard import it above?

lib/vsc/install/shared_setup.py
```diff
@@ -24,19 +24,46 @@
...
+try:
+    from unittest.suite import TestSuite
+except ImportError:
+    from unittest import TestSuite
+
 from distutils import log  # also for setuptools
 from distutils.dir_util import remove_tree
 
+from setuptools.command.test import test as TestCommand
+from setuptools.command.test import ScanningLoader
```
boegel: What if `setuptools` is not available? It's not a part of the standard Python library...

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+            if re.search(test_filter, t._testMethodName):
+                res.addTest(t)
+    return res
+
+
+class VscScanningLoader(ScanningLoader):
+    def loadTestsFromModule(self, module):
+        """Allow for filter"""
+        testsuites = ScanningLoader.loadTestsFromModule(self, module)
+        test_filter = getattr(__builtin__,'__test_filter')
```
boegel: space after `,`

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+    setattr(__builtin__, '__test_filter',  {
+        'module': None,
+        'function': None,
+        'allowmods': [],
+    })
+
+
+def filter_testsuites(testsuites):
+    """(Recursive) filtering of (suites of) tests"""
+    test_filter = getattr(__builtin__,'__test_filter')['function']
```
boegel: space after `,`

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+            res.addTest(filter_testsuites(t))
+        else:
+            if re.search(test_filter, t._testMethodName):
+                res.addTest(t)
+    return res
+
+
+class VscScanningLoader(ScanningLoader):
+    def loadTestsFromModule(self, module):
+        """Allow for filter"""
```
boegel: needs a better docstring, should provide a better description of what `loadTestsFromModule` does/returns?

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+        testsuites = ScanningLoader.loadTestsFromModule(self, module)
+        test_filter = getattr(__builtin__,'__test_filter')
+
+        res = testsuites
+
+        if test_filter['module'] is not None:
+            mod = module.__name__
```
boegel: avoid confusion between `mod` and `module`, use `modname` instead of `mod`?

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+    """
+    The cmdclass for testing
+    """
+    user_options = TestCommand.user_options + [
+        ('test-filterf=', 'f', "Regex filter on test function names"),
+        ('test-filterm=', 'F', "Regex filter on test (sub)modules"),
+    ]
+
+    def initialize_options(self):
+        TestCommand.initialize_options(self)
```
boegel: missing docstring

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+        self.test_filterm = None
+        self.test_filterf = None
+        self.test_loader = "vsc.install.shared_setup:VscScanningLoader"
+
+    def run_tests(self):
```
boegel: missing docstring

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+    user_options = TestCommand.user_options + [
+        ('test-filterf=', 'f', "Regex filter on test function names"),
+        ('test-filterm=', 'F', "Regex filter on test (sub)modules"),
+    ]
+
+    def initialize_options(self):
+        TestCommand.initialize_options(self)
+        self.test_filterm = None
+        self.test_filterf = None
+        self.test_loader = "vsc.install.shared_setup:VscScanningLoader"
```
boegel: magic string?

combine `VscScanningLoader.__name__` with whatever you need to use to determine the module name you're in?

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+
+    def run_tests(self):
+        getattr(__builtin__,'__test_filter').update({
+            'function': self.test_filterf,
+            'module': self.test_filterm,
+        })
+
+        # should be setup.py
+        setup_py = os.path.abspath(sys.argv[0])
```
boegel: you can verify whether it is, by checking `os.path.basename(setup_py)`?

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+        log.info('run_tests from %s' % setup_py)
+        base_dir = os.path.dirname(setup_py)
+
+        # make a lib dir to trick setup.py to package this properly
+        # and git ignore empty dirs, so recreate it if necessary
+        lib_dir = os.path.join(base_dir, 'lib')
+        if not os.path.exists(lib_dir):
+            os.mkdir(lib_dir)
```
boegel: no cleanup?

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+        loaded_vsc_modules = [mod for mod in sys.modules.keys() if mod == 'vsc' or mod.startswith('vsc.')]
+        reload_vsc_modules = []
+        for mod in loaded_vsc_modules:
+            if hasattr(sys.modules[mod], '__file__'):
+                # only actual modules
+                reload_vsc_modules.append(mod)
+            del(sys.modules[mod])
+        # reimport
+        for mod in reload_vsc_modules:
+            __import__(mod)
```
boegel: how do you know whether you're importing the right stuff here?

because of the sys.path hack at the top?

maybe check here whether things are being imported where you expect from (use `base_dir`)?

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +187,123 @@ def run(self):
...
+    """
+    user_options = TestCommand.user_options + [
+        ('test-filterf=', 'f', "Regex filter on test function names"),
+        ('test-filterm=', 'F', "Regex filter on test (sub)modules"),
+    ]
+
+    def initialize_options(self):
+        TestCommand.initialize_options(self)
+        self.test_filterm = None
+        self.test_filterf = None
```
boegel: what magic takes care of populating these if `-f` or `-F` are used?

lib/vsc/install/shared_setup.py
```diff
@@ -283,6 +415,11 @@ def action_target(target, setupfn=setup, extra_sdist=[]):
     if do_cleanup:
         cleanup()
 
+    if urltemplate:
+        target['url'] = urltemplate % target
+        if 'github' in urltemplate:
+            target['download_url'] = "%s/tarball/master" % target['url']
```
boegel: flesh out `master` into a `branch` named argument?

boegel:
@stdweird: fully reviewed, mostly minor remarks and things that are not fully clear (to me)

I still feel there's a lot of stuff being done here, which makes it feel like we're missing something, it's too complicated. Although, the filter support does add quite a bit the code in here, of course.

test/runner.py
stdweird: this is for people who run the tests with `python test/runner.py`

lib/vsc/install/shared_setup.py
stdweird: #195

lib/vsc/install/shared_setup.py
stdweird: no really, this is simply an empty directory that is missing because git doesn't handle empty dirs.
but i'll add the cleanup for good practise.

test/runner.py
boegel: Could you avoid this hack by using relative imports? `import .docs ad tdo`, etc?

stdweird: how would relative paths pickup something differently then what `sys.path` sets as order?

boegel: don't relative imports imply only looking in the dir where the current module is located?

not really sure here, trying to find alternatives, the hack just feels wrong (but I guess it's not a blocker)

lib/vsc/install/shared_setup.py
stdweird: the problem is i don't know which one of the vsc modules that should be loaded from the repo

boegel: hmm, right

test/runner.py
stdweird: the whole `runner.py` can be removed :smile:

boegel: ok, even better, but then we should bump to 2.4?

stdweird: ok

lib/vsc/install/shared_setup.py
stdweird: are you? (and are you suggestion emacs' sort-fields is wrong :smile: )

lib/vsc/install/shared_setup.py
stdweird: yes, py2.6 ism

lib/vsc/install/shared_setup.py
stdweird: i cleaned up this setuptools legacy code a bit

lib/vsc/install/shared_setup.py
stdweird: i added a proper comment

lib/vsc/install/shared_setup.py
stdweird: setuptools

lib/vsc/install/shared_setup.py
stdweird: ok

lib/vsc/install/shared_setup.py
stdweird: hmm, what if someone uses it from a `setup_whatever.py` to test branches?

stdweird:
retest this please

lib/vsc/install/shared_setup.py
boegel: some alphabet, different sorting key (variable name vs actual name of person), so nvm :)

lib/vsc/install/shared_setup.py
```diff
@@ -24,19 +24,70 @@
...
+# egg_info uses sdist directly through manifest_maker
+from setuptools.command.sdist import sdist
+from setuptools.command.egg_info import egg_info
+
+try:
+    # py2.7+
+    from unittest.suite import TestSuite
+except ImportError:
+    # py2.6
+    from unittest import TestSuite
```
boegel: no need for the exception for py2.7?

```
-bash-4.1$ python -c "import sys; print sys.version; from unittest import TestSuite"
2.6.6 (r266:84292, Jul 22 2015, 16:47:47) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-16)]
-bash-4.1$ module load Python/2.7.9-foss-2015a
-bash-4.1$ python -c "import sys; print sys.version; from unittest import TestSuite"
2.7.9 (default, Sep  1 2015, 15:59:22) 
[GCC 4.9.2]
-bash-4.1$ 
```

```
-bash-4.1$ python -c "import sys; print sys.version; from unittest.suite import TestSuite as T1; from unittest import TestSuite as T2; print T1 is T2"
2.7.9 (default, Sep  1 2015, 15:59:22) 
[GCC 4.9.2]
True
```

stdweird: ok, probably some __init__ magic

lib/vsc/install/shared_setup.py
boegel: On 09/10/15 09:48, Stijn De Weirdt wrote:
>
> In lib/vsc/install/shared_setup.py 
> <https://github.com/hpcugent/vsc-base/pull/193#discussion_r41604865>:
>
> > +    ]
> > +
> > +    def initialize_options(self):
> > +        TestCommand.initialize_options(self)
> > +        self.test_filterm = None
> > +        self.test_filterf = None
> > +        self.test_loader = "vsc.install.shared_setup:VscScanningLoader"
> > +
> > +    def run_tests(self):
> > +        getattr(__builtin__,'__test_filter').update({
> > +            'function': self.test_filterf,
> > +            'module': self.test_filterm,
> > +        })
> > +
> > +        # should be setup.py
> > +        setup_py = os.path.abspath(sys.argv[0])
>
> hmm, what if someone uses it from a |setup_whatever.py| to test branches?
>

Then the check will fail. You can make it a bit more flexible if you 
need to:

os.path.basename(setup_py).startswith('setup')
os.path.basename(setup_py).endswith('.py')

lib/vsc/install/shared_setup.py
```diff
@@ -24,19 +24,65 @@
...
 
+from setuptools.command.test import test as TestCommand
+from setuptools.command.test import ScanningLoader
+from setuptools import setup
+from setuptools.command.bdist_rpm import bdist_rpm
+from setuptools.command.build_py import build_py
+from setuptools.command.install_scripts import install_scripts
+# egg_info uses sdist directly through manifest_maker
+from setuptools.command.sdist import sdist
+from setuptools.command.egg_info import egg_info
```
boegel: sort these?

lib/vsc/install/shared_setup.py
```diff
@@ -61,80 +107,35 @@ def remove_extra_bdist_rpm_files():
...
-sdw = ('Stijn De Weirdt', 'stijn.deweirdt@ugent.be')
-wdp = ('Wouter Depypere', 'wouter.depypere@ugent.be')
-kw = ('Kenneth Waegeman', 'Kenneth.Waegeman@UGent.be')
-eh = ('Ewan Higgs', 'Ewan.Higgs@UGent.be')
+    All __init__.py files that augment namespace packages should be installed by the
+    dependent package, so we need not install it here.
+    """
 
+    def find_sources(self):
+        """Fins the sources as default and then drop the cruft."""
```
boegel: fix typo: Fins -> Finds

lib/vsc/install/shared_setup.py
```diff
@@ -61,80 +107,35 @@ def remove_extra_bdist_rpm_files():
...
+        vsc_egg_info.find_sources(self)
+        for f in remove_extra_bdist_rpm_files():
```
boegel: no single-letter variables outside of list comprehensions please

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +175,161 @@ def run(self):
...
+        return res
+
+
+class VscTestCommand(TestCommand):
+    """
+    The cmdclass for testing
+    """
+
+    # make 2 new 'python setup.py test' options available:
+    #    --test-filterf and --test-filertm
```
boegel: also mention -f and -F, since that how we'll be actually using them?

stdweird: or use `--help`?

lib/vsc/install/shared_setup.py
```diff
@@ -174,6 +175,161 @@ def run(self):
...
+        # make a lib dir to trick setup.py to package this properly
+        # and git ignore empty dirs, so recreate it if necessary
+        lib_dir = os.path.join(base_dir, 'lib')
+        if not os.path.exists(lib_dir):
+            os.mkdir(lib_dir)
+            cleanup.append(lib_dir)
+
+        test_dir = os.path.join(base_dir, DEFAULT_TEST_SUITE)
+        if os.path.isdir(test_dir):
+            sys.path.insert(0, test_dir)
```
boegel: wait, shouldn't this add `base_dir` rather than `test_dir` to `sys.path`?

lib/vsc/install/shared_setup.py
boegel: either you mention both long/short options here, or you don't mention them at all

lib/vsc/install/shared_setup.py
stdweird: no, should add testdir for modules in test dir that are no unittests but are used in unittests

lib/vsc/install/shared_setup.py
stdweird: ok

lib/vsc/install/shared_setup.py
boegel: ok

boegel:
Going in, let's see what's broken. ;-)

test/runner.py
```diff
@@ -1,32 +0,0 @@
...
-
-
-from vsc.utils import fancylogger
-fancylogger.logToScreen(enable=False)
-
-suite = unittest.TestSuite([x.suite() for x in (a, td, tg, tf, te, tm, trest, trun, tt, topt, wrapt, tdo)])
-
-try:
-    import xmlrunner
-    rs = xmlrunner.XMLTestRunner(output="test-reports").run(suite)
```
boegel: @stdweirdt: Jenkins relies on having this available

PR #196, opened 2015-10-09
Support --test-xmlrunner=test-reports for junit test report output

stdweird:


lib/vsc/install/shared_setup.py
```diff
@@ -304,6 +307,33 @@ def reload_vsc_modules(self):
...
 
+    def force_xmlrunner(self):
+        """
+        A monkey-patch attempt to run the tests with
+        xmlrunner.XMLTestRunner(output=xyz).run(suite)
+
+        E.g. in case of jenkins and you want junit compatible reports
+        """
+        import xmlrunner
+        import unittest
```
boegel: do we really want these imports here, not on top?

lib/vsc/install/shared_setup.py
```diff
@@ -304,6 +307,33 @@ def reload_vsc_modules(self):
...
+
+        xmlrunner_output = self.test_xmlrunner
+
+        class OutputXMLTestRunner(xmlrunner.XMLTestRunner):
+            """Force the output"""
+            def __init__(self, *args, **kwargs):
+                kwargs['output'] = xmlrunner_output
+                xmlrunner.XMLTestRunner.__init__(self,*args, **kwargs)
+
+        main_orig = unittest.main
```
boegel: separate with empty line?

also, use this instead?

```python
from unittest import main as unittest_main
```

lib/vsc/install/shared_setup.py
stdweird: unittest is ok, but xmlrunner would become a build_requires

lib/vsc/install/shared_setup.py
stdweird: no, i don't think you can use that for monley patching

lib/vsc/install/shared_setup.py
boegel: you can just ignore the `ImportError` for `xmlrunner` on top, and complain in here if it's not imported?

lib/vsc/install/shared_setup.py
boegel: `unittest_main` is just a reference to the original `unittest.main`, I don't see a difference with `main_orig` (other than the name ;-))

stdweird: ok then

stdweird: but now i need another import line. not doing that :smile:

boegel:
Thanks @stdweird!

PR #198, opened 2015-10-09
XmlRunner monkey patching should support recent versions of setuptools

stdweird:


lib/vsc/install/shared_setup.py
```diff
@@ -331,7 +332,15 @@ def __init__(self, *args, **kwargs):
                 kwargs['output'] = xmlrunner_output
                 xmlrunner.XMLTestRunner.__init__(self, *args, **kwargs)
 
-        main_orig = unittest.main
+        # old setuptools
+        main_name = 'main'
```
boegel: hmm, this is really `unittest.main`; does that work too?

```python
main_name = 'unittest.main'
```

lib/vsc/install/shared_setup.py
```diff
@@ -331,7 +332,15 @@ def __init__(self, *args, **kwargs):
...
+        main_orig = getattr(setuptools.command.test, main_name, None)
+        if main_orig is None:
+            # newer setuptools (e.g. 0.18)
+            main_name = 'unittest_main'
+            main_orig = getattr(setuptools.command.test, main_name, None)
+        if main_orig is None:
+            raise Exception('monkey patching XmlRunner failed')
```
boegel: this is almost copy-pasting, should be a loopy thing?

```python
cand_main_names = ['unittest.main', 'unittest_main']
for main_name in cand_main_names:
    main_orig = getattr(setuptools.command.test, main_name, None)
    if main_orig is not None:
        break
if main_orig is None:
    raise Exception("monkey patching XmlRunner failed")
```

lib/vsc/install/shared_setup.py
stdweird: i don't understand the remark. it's sufficient to monkey patch setuptools command test, not unittest per se

lib/vsc/install/shared_setup.py
stdweird: that's hardly more readble with only 2 values to test.

boegel: it's not about readability, it's about avoid duplicatie code, and making it trivial to extend when setuptools changes their mind in a future version (they will, trust me ;-))

lib/vsc/install/shared_setup.py
boegel: in older setuptools versions run_tests uses `unittest.main`, not `main`, so it's clearer to monkeypatch `setuptools.command.test.unittest.main`?

stdweird: the do use `main` too, (my 0.6.7 had it), not just `unittest.main`

stdweird: i added `main` to the list below :smile:

boegel:
Looks fine, let's see...  *drumroll*

PR #201, opened 2015-10-12
install.shared_setup: Do not reload the vsc modules by default

stdweird:


lib/vsc/install/shared_setup.py
```diff
@@ -315,7 +316,10 @@ def reload_vsc_modules(self):
 
         # reimport
         for name in reload_vsc_modules:
-            __import__(name)
+            try:
+                __import__(name)
+            except ImportError:
+                log.error('Failed to reload/import %s (sys.path %s). Continuing, fingers crossed.' % (name, sys.path))
```
boegel: hmm... :-)

shouldn't we prefer a hard crash here?

how do we know we're still testing what we're supposed to be testing if this fails?

stdweird: reraising the exception now

PR #206, opened 2015-10-19
fix setup.py, vsc-base (also) provides 'vsc' namespace

boegel:
this is required to have `vsc/__init__.py` installed with vsc-base...

not sure if anything else is required to avoid shipping `vsc/__init__.py` in `vsc-base` RPMs

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,10 +43,10 @@ VSC_INSTALL_REQ_VERSION = '0.9.0'
 
 PACKAGE = {
     'name': 'vsc-base',
-    'version': '2.4.9',
+    'version': '2.4.9.1',
     'author': [sdw, jt, ag, kh],
     'maintainer': [sdw, jt, ag, kh],
-    'packages': ['vsc.utils'],
+    'packages': ['vsc', 'vsc.utils'],
     'scripts': ['bin/logdaemon.py', 'bin/startlogdaemon.sh', 'bin/bdist_rpm.sh', 'bin/optcomplete.bash'],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -43,10 +43,10 @@
 
 PACKAGE = {
     'name': 'vsc-base',
-    'version': '2.4.9',
+    'version': '2.4.9.1',
```
stdweird: just bump to 2.4.10

setup.py
```diff
@@ -43,10 +43,10 @@
...
     'author': [sdw, jt, ag, kh],
     'maintainer': [sdw, jt, ag, kh],
-    'packages': ['vsc.utils'],
+    'packages': ['vsc', 'vsc.utils'],
```
stdweird: is now provided by vsc-install, it's one or the other.

setup.py
boegel: that'll screw up ordering, for example in `module avail`, sure you want `2.4.10`?

setup.py
boegel: OK, but `vsc` should be listed here (and in all other `vsc` packages).

`vsc-mympirun` has it too, for example: https://github.com/hpcugent/vsc-mympirun/blob/master/setup.py#L43

So does `vsc-utils`: https://github.com/hpcugent/vsc-utils/blob/master/setup.py#L57

setup.py
stdweird: it's the next version, if you care about mudles, we'll go 2.5.0

setup.py
stdweird: yes, i saw that too. i corrected my comment.

boegel:
superseded by #207

PR #209, opened 2015-10-28
bump version to 2.4.12 to fix faultly tarball that was pushed on PyPI

boegel:


boegel:
@stdweird: please merge ^_^

PR #217, opened 2016-02-03
Mpi4py condition

### LINKED ISSUE #216: import from mpi4py is causing python to crash on login nodes
The following mpi4py import on fancylogger is causing python to crash on login nodes (where we don't have a working MPI subsystem):
```
from mpi4py import MPI
```
The author of mpi4py suggested a couple of workarounds, the first being to precede the import with: 

```
import mpi4py.rc
mpi4py.rc.initialize = False
``` 

This would fix the crash on the login nodes, but I suppose the following call to `MPI.COMM_WORLD.Get_rank()` would stop working on the compute nodes.

For keeping track of the discussion, I'm also pasting here another proposed solution by the author of mpi4py: 

> However, you could write your code this way:
>
>from mpi4py import MPI
>if MPI.Is_initialized():
>    rank_id = MPI.COMM_WORLD.Get_rank()
>else:
>    rank_id = 0 # or None, and special case None values elsewhere in the logger code

Here we have some clues on how to fix the problem, however it is not clear to me how we can fix that in a way that it works on both compute and login nodes.

Any ideas?

stdweird:
Fixes #216 by introducing `FANCYLOGGER_IGNORE_MPI4PY` enviroment variable


stdweird:
@JensTimmerman clueless about the errors. i can reproduce the fancylogger one, but no clue where the generaloption one comes from (nor what it means; probably a reraise of a previous exception)

test/fancylogger.py
```diff
@@ -75,6 +75,8 @@ class FancyLoggerTest(TestCase):
     fd = None
 
     def setUp(self):
+        super(FancyLoggerTest, self).setUp()
```
JensTimmerman: why move this up? are you sure this has the correct side effects?

JensTimmerman:
do we still need these python 2.4 workarounds, like
```
    def format_help(self, formatter=None):
        """For py2.4 compatibility reasons (missing epilog). This is the py2.7 / optparse 1.5.3 code"""

```

test/fancylogger.py
stdweird: in the sense that it doesn't matter, it has the correct side effects. i'm still ruling out any possible causes.

ehiggs:
aiui, no we don't need the 2.4 workarounds.

lib/vsc/utils/generaloption.py
```diff
@@ -43,7 +43,7 @@
 from optparse import OptionParser, OptionGroup, Option, Values, HelpFormatter
 from optparse import BadOptionError, SUPPRESS_USAGE, NO_DEFAULT, OptionValueError
 from optparse import SUPPRESS_HELP as nohelp  # supported in optparse of python v2.4
-from optparse import _ as _gettext  # this is gettext normally
+from optparse import gettext as _gettext  # this is gettext.gettext normally
```
ehiggs: Why not pull this from gettext.gettext and alias it as `_` as [suggested in the docs](https://docs.python.org/2/library/gettext.html#gettext.gettext).

lib/vsc/utils/generaloption.py
```diff
@@ -119,7 +119,11 @@ def get_empty_add_flex(allvalues):
 
     if empty is None:
         msg = "get_empty_add_flex cannot determine empty element for type %s (%s)"
-        self.log.raiseException(msg % (type(allvalues), allvalues))
+        msg = msg % (type(allvalues), allvalues)
+        if self is None:
+            raise Exception(msg)
```
ehiggs: Is there no better exception type to use?

test/fancylogger.py
```diff
@@ -368,23 +369,29 @@ def test_normal_logging(self):
         """
         Test if just using import logging, logging.warning still works after importing fancylogger
         """
-        _stderr = sys.stderr
```
ehiggs: Doesn't removing this break `sys.stderr` for the rest of the suite?

lib/vsc/utils/generaloption.py
stdweird: any suggestions?

stdweird:
test this please

lib/vsc/utils/generaloption.py
boegel: ValueError? TypeError?

stdweird:
test this please

lib/vsc/utils/generaloption.py
JensTimmerman: create a `NoEmptyElementFlexException`

stdweird: no new exceptions, i'll use typeerror

lib/vsc/utils/generaloption.py
```diff
@@ -119,7 +119,11 @@ def get_empty_add_flex(allvalues):
...
+        else:
+            self.log.raiseException(msg)
```
ehiggs: Make this exception a consistent type with whatever's chosen above (TypeError, I guess). So `except TypeError` works.

lib/vsc/utils/generaloption.py
stdweird: because it have no clue what the added value is and only confuses things

test/fancylogger.py
stdweird: it's handled by the `super(FancyLoggerTest, self).setUp() / tearDown`

boegel:
@stdweird: so, EB should define `$FANCYLOGGER_IGNORE_MPI4PY` (since it makes no sense to poll `mpi4py` in the EB logger)?

PR #223, opened 2016-03-03
only include root logger name in non-debug mode

boegel:
`getRootLoggerName()` always returns "not available in optimized mode" when it is run under `python -O` (i.e., when `__debug__` is `False`), so it makes no sense to include it to construct the "full logger name"

### DIFF
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -357,7 +357,9 @@ def getLogger(name=None, fname=False, clsname=False, fancyrecord=None):
     FANCYLOG_FANCYRECORD to False, or will also be disabled if a Name is set (and fancyrecord and
     module constant FANCYLOG_FANCYRECORD are also not set).
     """
-    nameparts = [getRootLoggerName()]
+    nameparts = []
+    if __debug__:
+        nameparts.append(getRootLoggerName())
 
     if fancyrecord is None:
         # Altough we could set it as default value in the function definition
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ from vsc.install.shared_setup import ag, kh, jt, sdw
 VSC_INSTALL_REQ_VERSION = '0.9.19'
 
 PACKAGE = {
-    'version': '2.4.18',
+    'version': '2.4.19',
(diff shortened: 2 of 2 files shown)


stdweird:
what does this change?

stdweird:
sorry, didn't read the description :smile:

stdweird:
test this please

boegel:
In the case of EasyBuild (where `python -O` is used/`$PYTHONOPTIMIZE` is set), you get this as log messages:

```
== 2016-03-03 09:52:17,768 tools.robot INFO Using robot path(s): ['/Users/kehoste/work/easybuild-easyconfigs/easybuild/easyconfigs']
```

instead of this:

```
== 2016-03-03 09:52:17,768 not available in optimized mode.tools.robot INFO Using robot path(s): ['/Users/kehoste/work/easybuild-easyconfigs/easybuild/easyconfigs']
```

If `python -O` is not used, it doesn't change anything.

boegel:
Please hold off merging this a little bit, since this breaks some nasty bit of code `eb`:

```
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/Users/kehoste/work/easybuild-framework/easybuild/main.py", line 47, in <module>
    from easybuild.tools.build_log import EasyBuildError, init_logging, print_msg, print_error, stop_logging
  File "easybuild/tools/build_log.py", line 182, in <module>
    del _init_fancylog.manager.loggerDict[_init_fancylog.name]
KeyError: 'root'
```

boegel:
Jenkins: test this please

boegel:
OK, this is not going to work, apparently the root logger name is key, even under `python -O` when it's "not available in debug mode"...

I'll figure out another way to get rid of the "not available in debug mode" in the EB logs.

boegel:
superior solution in EB is to simply not include the logger name, but log location instead: https://github.com/hpcugent/easybuild-framework/pull/1649

PR #224, opened 2016-03-03
Hercules -> FWO

boegel:


itkovian:
Should Hercules not stay in there with a timespan? And then start the FWO from 2016xyuv?

boegel:
Hercules now is part of FWO, let's not try and keep track of Flemish politics history in our license headers?

Also, that's how we've updated other packages, and how vsc-install checks for it, cfr. https://github.com/hpcugent/vsc-install/pull/30

PR #225, opened 2016-03-25
Add support for colorization of log messages sent to the screen.

riccardomurri:
The support depends on the external modules `coloredlogs` and
`humanfriendly`; if they are not available, logs will not be colorized
no matter what.

An additional option to `fancylogger.logToScreen()` allows turning log
colorization depending on terminal capabilities, or on/off unconditionally.

**Note that this makes EasyBuild logs colored by default!**

### DIFF
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -543,7 +543,7 @@ def logToUDP(hostname, port=5005, enable=True, datagramhandler=None, name=None):
 
 
 def _logToSomething(handlerclass, handleropts, loggeroption,
-                    enable=True, name=None, handler=None, formatterclass=logging.Formatter):
+                    enable=True, name=None, handler=None, formatterclass=None):
     """
     internal function to enable (or disable) logging to handler named handlername
     handleropts is options dictionary passed to create the handler instance
@@ -553,6 +553,9 @@ def _logToSomething(handlerclass, handleropts, loggeroption,
     if you want to disable logging to the handler, pass the earlier obtained handler
     """
     logger = getLogger(name, fname=False, clsname=False)
+
+    if formatterclass is None:
+        formatterclass = logging.Formatter
 
     if not hasattr(logger, loggeroption):
         # not set.
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -682,8 +682,7 @@ def _screenLogFormatterFactory(colorize=Colorize.NEVER, stream=sys.stdout):
         elif colorize == Colorize.ALWAYS:
             formatter = coloredlogs.ColoredFormatter
         else:
-            assert colorize == Colorize.NEVER, \
-                "Argument `colorize` must be one of 'auto', 'always', or 'never'."
+            raise ValueError("Argument `colorize` must be one of 'auto', 'always', or 'never'.")
     return formatter
 
 
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -75,6 +75,7 @@ Logging to a udp server:
 @author: Kenneth Hoste (Ghent University)
 """
 
+from collections import namedtuple
 import inspect
 import logging
 import logging.handlers
@@ -175,11 +176,7 @@ BACKUPCOUNT = 10  # number of rotating log files to save
 DEFAULT_UDP_PORT = 5005
 
 # poor man's enum
-class Colorize:
-    """Allowed values for the `logToScreen`'s `colorize` parameter."""
-    AUTO = 'auto'
-    ALWAYS = 'always'
-    NEVER = 'never'
+Colorize = namedtuple('Colorize', 'AUTO ALWAYS NEVER')('auto', 'always', 'never')
 
 # register new loglevelname
 logging.addLevelName(logging.CRITICAL * 2 + 1, 'APOCALYPTIC')
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -442,7 +442,7 @@ def getRootLoggerName():
         return "not available in optimized mode"
 
 
-def logToScreen(enable=True, handler=None, name=None, stdout=False, color=None):
+def logToScreen(enable=True, handler=None, name=None, stdout=False, color=False):
     """
     enable (or disable) logging to screen
     returns the screenhandler (this can be used to later disable logging to screen)
@@ -458,27 +458,22 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False, color=None):
     The `color` parameter enables or disables log colorization using
     ANSI terminal escape sequences, according to the following values:
 
-    * when `color` is ``auto`` or ``None`` (default), then try to
+    * when `color` is ``auto``, then try to
       auto-detect whether the output stream is connected to a terminal;
     * when `color` is ``True`` or the string ``'yes'``, then turn on
       log colorization unconditionally,
-    * any other value turns off log colorization unconditionally.
+    * any other value turns off log colorization unconditionally (default).
     """
     handleropts = {'stdout': stdout}
 
-    use_color_formatter = False  # default
+    formatter = logging.Formatter  # default
     if HAVE_COLOREDLOGS_MODULE:
-        if color is None or color == 'auto':
+        if color == 'auto':
             # auto-detect
             if humanfriendly.terminal.terminal_supports_colors(sys.stdout if stdout else sys.stderr):
-                use_color_formatter = True
+                formatter = coloredlogs.ColoredFormatter
         elif color is True or color == 'yes':
-            use_color_formatter = True
-
-    if use_color_formatter:
-        formatter = coloredlogs.ColoredFormatter
-    else:
-        formatter = logging.Formatter
+            formatter = coloredlogs.ColoredFormatter
 
     return _logToSomething(FancyStreamHandler,
                            handleropts,
@@ -486,7 +481,7 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False, color=None):
                            name=name,
                            enable=enable,
                            handler=handler,
-                           formatterclass=formatter
+                           formatterclass=formatter,
                            )
 
 
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -463,7 +463,7 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False, colorize='ne
     (which see).
     """
     handleropts = {'stdout': stdout}
-    formatter = _screenLogFormatterFactory(colorize=color, stream=(sys.stdout if stdout else sys.stderr))
+    formatter = _screenLogFormatterFactory(colorize=colorizec, stream=(sys.stdout if stdout else sys.stderr))
 
     return _logToSomething(FancyStreamHandler,
                            handleropts,
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -85,12 +85,14 @@ import traceback
 import weakref
 from distutils.version import LooseVersion
 
-try:
-    import coloredlogs
-    import humanfriendly
-    HAVE_COLOREDLOGS_MODULE = True
-except ImportError:
-    HAVE_COLOREDLOGS_MODULE = False
+HAVE_COLOREDLOGS_MODULE = False
+if os.environ.get('FANCYLOGGER_NO_COLOREDLOGS', '0').lower() not in ('1', 'yes', 'true', 'y'):
+    try:
+        import coloredlogs
+        import humanfriendly
+        HAVE_COLOREDLOGS_MODULE = True
+    except ImportError:
+        pass
 
 # constants
 TEST_LOGGING_FORMAT = '%(levelname)-10s %(name)-15s %(threadName)-10s  %(message)s'
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -463,7 +463,7 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False, colorize='ne
     (which see).
     """
     handleropts = {'stdout': stdout}
-    formatter = _screenLogFormatterFactory(colorize=colorizec, stream=(sys.stdout if stdout else sys.stderr))
+    formatter = _screenLogFormatterFactory(colorize=colorize, stream=(sys.stdout if stdout else sys.stderr))
 
     return _logToSomething(FancyStreamHandler,
                            handleropts,
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -444,7 +444,7 @@ def getRootLoggerName():
         return "not available in optimized mode"
 
 
-def logToScreen(enable=True, handler=None, name=None, stdout=False, color=False):
+def logToScreen(enable=True, handler=None, name=None, stdout=False, color='never'):
     """
     enable (or disable) logging to screen
     returns the screenhandler (this can be used to later disable logging to screen)
@@ -462,20 +462,12 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False, color=False)
 
     * when `color` is ``auto``, then try to
       auto-detect whether the output stream is connected to a terminal;
-    * when `color` is ``True`` or the string ``'yes'``, then turn on
+    * when `color` is the string ``'always'``, then turn on
       log colorization unconditionally,
     * any other value turns off log colorization unconditionally (default).
     """
     handleropts = {'stdout': stdout}
-
-    formatter = logging.Formatter  # default
-    if HAVE_COLOREDLOGS_MODULE:
-        if color == 'auto':
-            # auto-detect
-            if humanfriendly.terminal.terminal_supports_colors(sys.stdout if stdout else sys.stderr):
-                formatter = coloredlogs.ColoredFormatter
-        elif color is True or color == 'yes':
-            formatter = coloredlogs.ColoredFormatter
+    formatter = _getScreenLogFormatter((sys.stdout if stdout else sys.stderr), color)
 
     return _logToSomething(FancyStreamHandler,
                            handleropts,
@@ -596,6 +588,28 @@ def _logToSomething(handlerclass, handleropts, loggeroption,
             logger.removeHandler(handler)
         setattr(logger, loggeroption, False)
     return handler
+
+
+def _getScreenLogFormatter(stream, colorize='never'):
+    """
+    Return a log formatter, with optional colorization features.
+
+    Second argument `colorize` controls whether the formatter
+    can use ANSI terminal escape sequences:
+
+    * ``'never'`` (default) forces use of the plain `logging.Formatter` class;
+    * ``'always'`` forces use of the colorizing formatter;
+    * ``'auto'`` selects the colorizing formatter depending on whether `stream` is connected to a terminal.
+    """
+    formatter = logging.Formatter  # default
+    if HAVE_COLOREDLOGS_MODULE:
+        if colorize == 'auto':
+            # auto-detect
+            if humanfriendly.terminal.terminal_supports_colors(stream):
+                formatter = coloredlogs.ColoredFormatter
+        elif colorize == 'always':
+            formatter = coloredlogs.ColoredFormatter
+    return formatter
 
 
 def _getSysLogFacility(name=None):
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -86,7 +86,7 @@ import weakref
 from distutils.version import LooseVersion
 
 
-def _disabled_by_environ(varname, default=False):
+def _disabled_by_environ(varname, default=True):
     """
     Compute a boolean based on the truth value of environment variable `varname`.
     If no variable by that name is present in `os.environ`, then return `default`.
@@ -137,11 +137,11 @@ def _disabled_by_environ(varname, default=False):
       42
 
     By default, calling `_disabled_by_environ` on an undefined
-    variable returns Python ``False``::
+    variable returns Python ``True``::
 
       >>> if 'NO_FOOBAR' in os.environ: del os.environ['NO_FOOBAR']
       >>> _disabled_by_environ('NO_FOOBAR')
-      False
+      True
     """
     if varname not in os.environ:
         return default
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -681,6 +681,8 @@ def _screenLogFormatterFactory(colorize=Colorize.NEVER, stream=sys.stdout):
                 formatter = coloredlogs.ColoredFormatter
         elif colorize == Colorize.ALWAYS:
             formatter = coloredlogs.ColoredFormatter
+        elif colorize == Colorize.NEVER:
+            pass
         else:
             raise ValueError("Argument `colorize` must be one of 'auto', 'always', or 'never'.")
     return formatter
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -85,8 +85,72 @@ import traceback
 import weakref
 from distutils.version import LooseVersion
 
+
+def _disabled_by_environ(varname, default=False):
+    """
+    Compute a boolean based on the truth value of environment variable `varname`.
+    If no variable by that name is present in `os.environ`, then return `default`.
+
+    For the purpose of this function, the string values ``'1'``,
+    ``'y'``, ``'yes'``, and ``'true'`` (case-insensitive) are all
+    mapped to the truth value ``True``::
+
+      >>> os.environ['NO_FOOBAR'] = '1'
+      >>> _disabled_by_environ('NO_FOOBAR')
+      True
+      >>> os.environ['NO_FOOBAR'] = 'Y'
+      >>> _disabled_by_environ('NO_FOOBAR')
+      True
+      >>> os.environ['NO_FOOBAR'] = 'Yes'
+      >>> _disabled_by_environ('NO_FOOBAR')
+      True
+      >>> os.environ['NO_FOOBAR'] = 'yes'
+      >>> _disabled_by_environ('NO_FOOBAR')
+      True
+      >>> os.environ['NO_FOOBAR'] = 'True'
+      >>> _disabled_by_environ('NO_FOOBAR')
+      True
+      >>> os.environ['NO_FOOBAR'] = 'TRUE'
+      >>> _disabled_by_environ('NO_FOOBAR')
+      True
+      >>> os.environ['NO_FOOBAR'] = 'true'
+      >>> _disabled_by_environ('NO_FOOBAR')
+      True
+
+    Any other value is mapped to Python ``False``::
+
+      >>> os.environ['NO_FOOBAR'] = '0'
+      >>> _disabled_by_environ('NO_FOOBAR')
+      False
+      >>> os.environ['NO_FOOBAR'] = 'no'
+      >>> _disabled_by_environ('NO_FOOBAR')
+      False
+      >>> os.environ['NO_FOOBAR'] = 'if you please'
+      >>> _disabled_by_environ('NO_FOOBAR')
+      False
+
+    If no variable named `varname` is present in `os.environ`, then
+    return `default`::
+
+      >>> del os.environ['NO_FOOBAR']
+      >>> _disabled_by_environ('NO_FOOBAR', 42)
+      42
+
+    By default, calling `_disabled_by_environ` on an undefined
+    variable returns Python ``False``::
+
+      >>> if 'NO_FOOBAR' in os.environ: del os.environ['NO_FOOBAR']
+      >>> _disabled_by_environ('NO_FOOBAR')
+      False
+    """
+    if varname not in os.environ:
+        return default
+    else:
+        return os.environ.get(varname).lower() in ('1', 'yes', 'true', 'y')
+
+
 HAVE_COLOREDLOGS_MODULE = False
-if os.environ.get('FANCYLOGGER_NO_COLOREDLOGS', '0').lower() not in ('1', 'yes', 'true', 'y'):
+if not _disabled_by_environ('FANCYLOGGER_NO_COLOREDLOGS'):
     try:
         import coloredlogs
         import humanfriendly
@@ -120,7 +184,7 @@ logging._levelNames['QUIET'] = logging.WARNING
 
 # mpi rank support
 _MPIRANK = MPIRANK_NO_MPI
-if os.environ.get('FANCYLOGGER_IGNORE_MPI4PY', '0').lower() not in ('1', 'yes', 'true', 'y'):
+if not _disabled_by_environ('FANCYLOGGER_IGNORE_MPI4PY'):
     try:
         from mpi4py import MPI
         if MPI.Is_initialized():
@@ -392,7 +456,7 @@ def getLogger(name=None, fname=False, clsname=False, fancyrecord=None):
 
     l = logging.getLogger(fullname)
     l.fancyrecord = fancyrecord
-    if os.environ.get('FANCYLOGGER_GETLOGGER_DEBUG', '0').lower() in ('1', 'yes', 'true', 'y'):
+    if not _disabled_by_environ('FANCYLOGGER_GETLOGGER_DEBUG'):
         print 'FANCYLOGGER_GETLOGGER_DEBUG',
         print 'name', name, 'fname', fname, 'fullname', fullname,
         print "getRootLoggerName: ", getRootLoggerName()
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -44,7 +44,11 @@
     'maintainer': [sdw, jt, ag, kh],
     # as long as 1.0.0 is not out, vsc-base should still provide vsc.fancylogger
     # setuptools must become a requirement for shared namespaces if vsc-install is removed as requirement
-    'install_requires': ['vsc-install >= %s' % VSC_INSTALL_REQ_VERSION],
+    'install_requires': [
+        'vsc-install >= %s' % VSC_INSTALL_REQ_VERSION,
+        'coloredlogs',     # automatic log colorizer
```
stdweird: need to check if this ends up in the rpm dependencies or not. but i don not see the point of listing them here since they are conditional. setuptools has something for such conditional behaviour i think
@JensTimmerman do you  rememebr the required magic for optional requirements?

lib/vsc/utils/fancylogger.py
```diff
@@ -85,6 +85,13 @@
 import weakref
 from distutils.version import LooseVersion
 
+try:
```
stdweird: can you implement this like the `mpiy4py` check below? (i.e. set it to false by default, and allow an environment variable to control the import)

lib/vsc/utils/fancylogger.py
```diff
@@ -443,15 +450,39 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False):
 
     by default, logToScreen will log to stderr; logging to stdout instead can be done
     by setting the 'stdout' parameter to True
+
+    The `color` parameter enables or disables log colorization using
+    ANSI terminal escape sequences, according to the following values:
+
+    * when `color` is ``auto`` or ``None`` (default), then try to
+      auto-detect whether the output stream is connected to a terminal;
+    * when `color` is ``True`` or the string ``'yes'``, then turn on
```
stdweird: please make up if `color` is a string or not. do not mix types.

lib/vsc/utils/fancylogger.py
```diff
@@ -443,15 +450,39 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False):
...
+      log colorization unconditionally,
+    * any other value turns off log colorization unconditionally.
     """
     handleropts = {'stdout': stdout}
 
+    use_color_formatter = False  # default
```
stdweird: there is no point in the vairable, just start with the default formatter class, and change it if necessary

lib/vsc/utils/fancylogger.py
```diff
@@ -443,15 +450,39 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False):
...
+    else:
+        formatter = logging.Formatter
+
     return _logToSomething(FancyStreamHandler,
                            handleropts,
                            loggeroption='logtoscreen_stdout_%s' % str(stdout),
                            name=name,
                            enable=enable,
                            handler=handler,
+                           formatterclass=formatter
```
stdweird: add a `,` at the end

lib/vsc/utils/fancylogger.py
```diff
@@ -443,15 +450,39 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False):
...
+      log colorization unconditionally,
+    * any other value turns off log colorization unconditionally.
     """
     handleropts = {'stdout': stdout}
 
+    use_color_formatter = False  # default
+    if HAVE_COLOREDLOGS_MODULE:
+        if color is None or color == 'auto':
+            # auto-detect
+            if humanfriendly.terminal.terminal_supports_colors(sys.stdout if stdout else sys.stderr):
```
stdweird: i would move this check to the import above, maybe use a tuple `COLOREDLOGS` with the value of HAVE_COLOREDLOGS_MODULE and as second element the meaning of auto.

lib/vsc/utils/fancylogger.py
```diff
@@ -443,15 +450,39 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False):
...
+      auto-detect whether the output stream is connected to a terminal;
+    * when `color` is ``True`` or the string ``'yes'``, then turn on
+      log colorization unconditionally,
+    * any other value turns off log colorization unconditionally.
     """
     handleropts = {'stdout': stdout}
 
+    use_color_formatter = False  # default
+    if HAVE_COLOREDLOGS_MODULE:
+        if color is None or color == 'auto':
```
stdweird: the default shouldn't do anything, colored modules available or not. the current default means that as soon as the modules are present, the stdout logs are colorised.

lib/vsc/utils/fancylogger.py
```diff
@@ -512,7 +543,8 @@ def logToUDP(hostname, port=5005, enable=True, datagramhandler=None, name=None):
                            )
 
 
-def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=None, handler=None):
+def _logToSomething(handlerclass, handleropts, loggeroption,
+                    enable=True, name=None, handler=None, formatterclass=logging.Formatter):
```
stdweird: i would have to check, but the default formatterclass should be `formatterclass=None`, and then in the method
```python
if formatterclass is None:
    formatterclass=logging.Formatter
```

stdweird:
i think it would be better if you can pass the formatterclass via all logToX and to `_logToSomething`. then you can pass in EB whatever formatter you want making it is much more generic.

boegel:
Jenkins: ok to test

lib/vsc/utils/fancylogger.py
riccardomurri: Easy to do, but why?  if the import succeeds, you can use the feature, otherwise it will be ignored. It does not have side-effects like probing for `mpi4py` can have...

stdweird: that's exactly what we tought of the original `mpi4py` code :smile:

lib/vsc/utils/fancylogger.py
riccardomurri: Well, since we do not have a natural way of doing a Yes/No/Maybe enum in Python, I thought that 'auto'/'yes'/'no' would do, but then True/False makes also a lot of sense from an API point of view.

I do not have a strong opinion here, so just suggest what triple of values you would like to use and I'll make the function accept that.

lib/vsc/utils/fancylogger.py
riccardomurri: Well, it's possible that `logToScreen()` never gets called, or it's called with color="no" -- why performing the check then?

lib/vsc/utils/fancylogger.py
riccardomurri: Why? The only difference would be in the case where somebody is reassigning or moneky-patching `logging.Formatter

riccardomurri:
> i think it would be better if you can pass the formatterclass via
> all logToX and to _logToSomething. then you can pass in EB whatever
> formatter you want making it is much more generic.

It would be more generic but when are you really going to write ANSI
terminal escape sequences to a file?

lib/vsc/utils/fancylogger.py
stdweird: it's just some akward code in the middle of logToScreen. maybe move the whole color logic to a function that returns the colored formatter or None ?

lib/vsc/utils/fancylogger.py
stdweird: sure, but don't mix types. string/None is perfectly fine

stdweird:
@riccardomurri who knows why people want to change the formatter, and they can also change to any formatter, not just the colored one (i also don't know why people would use a colored formatter :smile: )

lib/vsc/utils/fancylogger.py
```diff
@@ -431,7 +438,7 @@ def getRootLoggerName():
         return "not available in optimized mode"
 
 
-def logToScreen(enable=True, handler=None, name=None, stdout=False):
+def logToScreen(enable=True, handler=None, name=None, stdout=False, color=False):
```
stdweird: what was  wrong with `None`? you're again mixing booleans and strings

lib/vsc/utils/fancylogger.py
stdweird: indeed.

lib/vsc/utils/fancylogger.py
riccardomurri: Just tell me what you would like to use: 'auto'/'yes'/'no' (all strings)? True/False/None? 0/1/2? named constants?

boegel: Since there are three cases, I would go with strings:

* `always`: always use color, even when the output device is not a terminal
* `yes` (or `auto`): color (only for terminal output device)
* `no`: no color

stdweird: a single type (and None).
so either strings/None or booleans/None. with boolean/None you are a bit limited, so True=always color/False=never color/None=auto is possible, but there' can't be a 4th option in the future.
if you use strings, None could be set equal to 'no/never'

boegel:
@riccardomurri we're missing a version bump in `setup.py`

lib/vsc/utils/fancylogger.py
```diff
@@ -443,8 +452,18 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False):
 
     by default, logToScreen will log to stderr; logging to stdout instead can be done
     by setting the 'stdout' parameter to True
+
+    The `color` parameter enables or disables log colorization using
```
stdweird: `enables or disables` -> `controls`

lib/vsc/utils/fancylogger.py
```diff
@@ -443,8 +452,18 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False):
...
+    ANSI terminal escape sequences, according to the following values:
+
+    * when `color` is ``auto``, then try to
+      auto-detect whether the output stream is connected to a terminal;
+    * when `color` is the string ``'always'``, then turn on
```
stdweird: remove `the string`

lib/vsc/utils/fancylogger.py
stdweird: better to add this piece of costring to the function, and say that the `color` parameter is passed to that function.

lib/vsc/utils/fancylogger.py
```diff
@@ -512,7 +532,8 @@ def logToUDP(hostname, port=5005, enable=True, datagramhandler=None, name=None):
                            )
 
 
-def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=None, handler=None):
+def _logToSomething(handlerclass, handleropts, loggeroption,
+                    enable=True, name=None, handler=None, formatterclass=None):
     """
     internal function to enable (or disable) logging to handler named handlername
     handleropts is options dictionary passed to create the handler instance
```
stdweird: add `formatterclass` to the docstring

lib/vsc/utils/fancylogger.py
```diff
@@ -562,6 +586,28 @@ def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=N
     return handler
 
 
+def _getScreenLogFormatter(stream, colorize='never'):
+    """
+    Return a log formatter, with optional colorization features.
```
stdweird: `formatter` -> `formatter class` (i.e. not an instance)

lib/vsc/utils/fancylogger.py
```diff
@@ -562,6 +586,28 @@ def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=N
     return handler
 
 
+def _getScreenLogFormatter(stream, colorize='never'):
```
stdweird: it's ok to call it `FormatterFactory`.

lib/vsc/utils/fancylogger.py
```diff
@@ -562,6 +586,28 @@ def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=N
...
+    """
+    Return a log formatter, with optional colorization features.
+
+    Second argument `colorize` controls whether the formatter
```
stdweird: and the first mandatory argument?

lib/vsc/utils/fancylogger.py
```diff
@@ -562,6 +586,28 @@ def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=N
...
+    can use ANSI terminal escape sequences:
+
+    * ``'never'`` (default) forces use of the plain `logging.Formatter` class;
```
stdweird: `forces` -> `returns`

lib/vsc/utils/fancylogger.py
```diff
@@ -562,6 +586,28 @@ def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=N
...
+    * ``'always'`` forces use of the colorizing formatter;
+    * ``'auto'`` selects the colorizing formatter depending on whether `stream` is connected to a terminal.
```
stdweird: so auto+redirected stdout should not be colorised?

ehiggs:
Would it not be better to vendorise the dependent packages here so 1) they're always available (and we can always have colours, yay!) and 2) reduce the bug vector surface in the face of competing packages by having consistent behaviour for a 'core' package.

boegel:
@ehiggs vendorize?

riccardomurri:
> @riccardomurri we're missing a version bump in `setup.py`

To what version should it bump to?

lib/vsc/utils/fancylogger.py
riccardomurri: > > +    Second argument `colorize` controls whether the formatter
> > +    can use ANSI terminal escape sequences:
> > +
> > +    * ``'never'`` (default) forces use of the plain `logging.Formatter` class;
> > +    * ``'always'`` forces use of the colorizing formatter;
> > +    * ``'auto'`` selects the colorizing formatter depending on whether `stream` is connected to a terminal.
>
> so auto+redirected stdout should not be colorised?

No. That's the whole point of having `auto`.

lib/vsc/utils/fancylogger.py
riccardomurri: > > +    Second argument `colorize` controls whether the formatter
> > +    can use ANSI terminal escape sequences:
> > +
> > +    * ``'never'`` (default) forces use of the plain `logging.Formatter` class;
>
> `forces` -> `returns`

Hmm... I kind of disagree: both are of course correct, but "force" is
there to stress that no check is done whether outputting ANSI control
sequences to `stream` is sensible or not.

ehiggs:
@boegel: bring the contents of the dependent packages in tree.

boegel:
@ehiggs but then we'll also need to maintain them, backport bug fixes, etc.?

Doesn't seem like a good idea to me.

@riccardomurri I don't think the warrants a bump to 2.5.0, so just a minor bump to 2.4.19?

setup.py
```diff
@@ -39,12 +39,18 @@
 VSC_INSTALL_REQ_VERSION = '0.9.19'
 
 PACKAGE = {
-    'version': '2.4.18',
+    'version': '2.5.0',
```
boegel: not sure if there's enough here to go to 2.5.0, but on the other hand, I don't really mind (the version number is getting quite weird anyway ;-))

lib/vsc/utils/fancylogger.py
```diff
@@ -443,8 +452,14 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False):
...
     by default, logToScreen will log to stderr; logging to stdout instead can be done
     by setting the 'stdout' parameter to True
+
+    The `color` parameter enables or disables log colorization using
+    ANSI terminal escape sequences, according to the values allowed
+    in the `colorize` parameter to function `_screenLogFormatterFactory`
+    (which see).
     """
     handleropts = {'stdout': stdout}
+    formatter = _screenLogFormatterFactory(color, sys.stdout if stdout else sys.stderr)
```
boegel: use named arguments?

```python
formatter = _screenLogFormatterFactory(colorize=colorize, stream=sys.stdout if stdout else sys.stderr)
```

lib/vsc/utils/fancylogger.py
boegel: +1, and I would rename it to `colorize` like in `_screenLogFormatterFactory`?

lib/vsc/utils/fancylogger.py
```diff
@@ -85,6 +85,15 @@
 import weakref
 from distutils.version import LooseVersion
 
+HAVE_COLOREDLOGS_MODULE = False
+if os.environ.get('FANCYLOGGER_NO_COLOREDLOGS', '0').lower() not in ('1', 'yes', 'true', 'y'):
```
boegel: Aren't we using the `('1', 'yes', 'true', 'y')` in multiple places now?

In that case, this should become a constant like `ENABLE_VALUES`?

lib/vsc/utils/fancylogger.py
```diff
@@ -562,6 +583,30 @@ def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=N
...
+    Second argument `stream` is the stream to check in case `colorize` is ``'auto'``.
+    """
+    formatter = logging.Formatter  # default
+    if HAVE_COLOREDLOGS_MODULE:
+        if colorize == 'auto':
+            # auto-detect
+            if humanfriendly.terminal.terminal_supports_colors(stream):
+                formatter = coloredlogs.ColoredFormatter
+        elif colorize == 'always':
+            formatter = coloredlogs.ColoredFormatter
```
boegel: do we care about other values than `auto`, `always` or `never` being passed down?

if so, this is missing:

```python
elif colorize != 'never':
    # FAIL
```

lib/vsc/utils/fancylogger.py
```diff
@@ -512,10 +528,12 @@ def logToUDP(hostname, port=5005, enable=True, datagramhandler=None, name=None):
...
 
 
-def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=None, handler=None):
+def _logToSomething(handlerclass, handleropts, loggeroption,
+                    enable=True, name=None, handler=None, formatterclass=None):
     """
     internal function to enable (or disable) logging to handler named handlername
-    handleropts is options dictionary passed to create the handler instance
+    handleropts is options dictionary passed to create the handler instance;
+    `formatterclass` is the class to use to instanciate a log formatter object.
```
boegel: instantiate? (`c` -> `t`)

ehiggs:
Repinning a git submodule takes almost no effort.
On 25 Mar 2016 11:49 p.m., Kenneth Hoste <notifications@github.com> wrote:@ehiggs but then we'll also need to maintain them, backport bug fixes, etc.?

Doesn't seem like a good idea to me.

@riccardomurri I don't think the warrants a bump to 2.5.0, so just a minor bump to 2.4.19?

riccardomurri:
> @riccardomurri I don't think the warrants a bump to 2.5.0, so just a
> minor bump to 2.4.19?

I presumed you were using [semantic versioning][1] so I bumped it to 2.5.0.

I think it's best if I take versioning out anyway -- deciding version
numbers is a task for the release manager, not random contributors :-)

[1]: http://semver.org/

lib/vsc/utils/fancylogger.py
riccardomurri: > > +        if colorize == 'auto':
> > +            # auto-detect
> > +            if humanfriendly.terminal.terminal_supports_colors(stream):
> > +                formatter = coloredlogs.ColoredFormatter
> > +        elif colorize == 'always':
> > +            formatter = coloredlogs.ColoredFormatter
>
> do we care about other values than `auto`, `always` or `never` being passed down?
>
> if so, this is missing:
>
> ```python
> elif colorize != 'never':
>     # FAIL
> ```
>

Done, except I've used an `assert` statement.

lib/vsc/utils/fancylogger.py
riccardomurri: > Aren't we using the `('1', 'yes', 'true', 'y')` in multiple places now?
>
> In that case, this should become a constant like `ENABLE_VALUES`?

+1 but out of scope for this PR ;-)

lib/vsc/utils/fancylogger.py
```diff
@@ -562,6 +583,32 @@ def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=N
...
+    formatter = logging.Formatter  # default
+    if HAVE_COLOREDLOGS_MODULE:
+        if colorize == 'auto':
+            # auto-detect
+            if humanfriendly.terminal.terminal_supports_colors(stream):
+                formatter = coloredlogs.ColoredFormatter
+        elif colorize == 'always':
+            formatter = coloredlogs.ColoredFormatter
+        else:
+            assert colorize == 'never', "Argument `colorize` must be one of 'auto', 'always', or 'never'."
```
boegel: we usually don't use `assert`, but here it makes sense imho, since this is basically a sanity check

lib/vsc/utils/fancylogger.py
```diff
@@ -443,8 +452,14 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False):
 
     by default, logToScreen will log to stderr; logging to stdout instead can be done
     by setting the 'stdout' parameter to True
+
+    The `colorize` parameter enables or disables log colorization using
+    ANSI terminal escape sequences, according to the values allowed
+    in the `colorize` parameter to function `_screenLogFormatterFactory`
+    (which see).
```
boegel: which see?

lib/vsc/utils/fancylogger.py
```diff
@@ -431,7 +440,7 @@ def getRootLoggerName():
         return "not available in optimized mode"
 
 
-def logToScreen(enable=True, handler=None, name=None, stdout=False):
+def logToScreen(enable=True, handler=None, name=None, stdout=False, colorize='never'):
```
boegel: does it make sense to use `None` here as default, and then fall back to using `'never'` in `_screenLogFormatterFactory` if it's `None`?

lib/vsc/utils/fancylogger.py
boegel: well, it's not like it's a lot of work? I wouldn't mind if that's included in this PR, since the changes done here are 'expoising' the issue...

boegel:
@stdweird any further thoughts on this? Other than a handful of minor things, this looks good to me...

Maybe enhancing the unit tests? Although that will get messy (w.r.t. required deps, etc.)

lib/vsc/utils/fancylogger.py
riccardomurri: > > +    The `colorize` parameter enables or disables log colorization using
> > +    ANSI terminal escape sequences, according to the values allowed
> > +    in the `colorize` parameter to function `_screenLogFormatterFactory`
> > +    (which see).
>
> which see?

"""
The abbreviation q.v. stands for quod vide, which translates literally
as “which see,” although in practice it means something more like “for
which see elsewhere.” It is used in notes after a word or phrase to
indicate that more information can be found about the topic somewhere
else in the current work.
"""

From:
http://writingcenter.unc.edu/handouts/latin-terms-and-abbreviations/
(which see ;-))

lib/vsc/utils/fancylogger.py
riccardomurri: > > -def logToScreen(enable=True, handler=None, name=None, stdout=False):
> > +def logToScreen(enable=True, handler=None, name=None, stdout=False, colorize='never'):
>
> does it make sense to use `None` here as default, and then fall back to using `'never'` in `_screenLogFormatterFactory` if it's `None`?

why? what advantages/differences would it have compared to defaulting to ``'never'``?

boegel: Just that the magic string `'never'` is spread out in multiple places; it's harder to mistype `None` without noticing in a decent editor.

JensTimmerman: use constants for these magic strings please, even beter then None

riccardomurri: > > -def logToScreen(enable=True, handler=None, name=None, stdout=False):
> > +def logToScreen(enable=True, handler=None, name=None, stdout=False, colorize='never'):
>
> use constants for these magic strings please, even beter then None

Then it would make sense to actually use a proper enum.

But it's another additional dependency package, are you sure you want
that?

JensTimmerman: well no, 
the point of vsc-base is that it doesn't have any external dependencies, 
things that have dependencies go into `vsc-utils`

riccardomurri:
Hello @stdweird and @boegel ! Any chance of having this merged soon-ish?

Ciao,
R

lib/vsc/utils/fancylogger.py
```diff
@@ -101,6 +174,13 @@
 
 DEFAULT_UDP_PORT = 5005
 
+# poor man's enum
```
stdweird: use namedtuple for this

lib/vsc/utils/fancylogger.py
```diff
@@ -85,6 +85,79 @@
...
+def _env_to_boolean(varname, default=False):
+    """
+    Compute a boolean based on the truth value of environment variable `varname`.
+    If no variable by that name is present in `os.environ`, then return `default`.
+
+    For the purpose of this function, the string values ``'1'``,
+    ``'y'``, ``'yes'``, and ``'true'`` (case-insensitive) are all
+    mapped to the truth value ``True``::
+
+      >>> os.environ['NO_FOOBAR'] = '1'
```
stdweird: is this picked up by the tests?

lib/vsc/utils/fancylogger.py
```diff
@@ -562,6 +654,35 @@ def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=N
...
+    formatter = logging.Formatter  # default
+    if HAVE_COLOREDLOGS_MODULE:
+        if colorize == Colorize.AUTO:
+            # auto-detect
+            if humanfriendly.terminal.terminal_supports_colors(stream):
+                formatter = coloredlogs.ColoredFormatter
+        elif colorize == Colorize.ALWAYS:
+            formatter = coloredlogs.ColoredFormatter
+        else:
+            assert colorize == Colorize.NEVER, \
```
stdweird: just raise here, to avoid problmes with `-O`

lib/vsc/utils/fancylogger.py
```diff
@@ -562,6 +654,35 @@ def _logToSomething(handlerclass, handleropts, loggeroption, enable=True, name=N
     return handler
 
 
+def _screenLogFormatterFactory(colorize=Colorize.NEVER, stream=sys.stdout):
```
stdweird: add unittest for this function, we'll add the required modules to the test enviroment

lib/vsc/utils/fancylogger.py
riccardomurri: ???? 

How can you simulate a "enum" (algebraic "sum" type) with a namedtuple (algebraic "product" type)?

lib/vsc/utils/fancylogger.py
riccardomurri: I will change it if you insist, but that's the whole point of using `assert`: that you do not incur in the performance penalty of checking when running code in production with `-O`.  Testing for interface compliance is done in "debug" mode, production should assume that the code is correct.

setup.py
riccardomurri: I assumed you were using [semantic versioning][1].  Just tell me what version number you would like to have here.

[1]: http://semver.org/

lib/vsc/utils/fancylogger.py
riccardomurri: It's a doctest.  Does your testing framework run doctests?

setup.py
boegel: Let's go with 2.5.0, no need to have a deep philosophical discussion about it. :)

lib/vsc/utils/fancylogger.py
boegel: There should be no performane penalty, since the `else` will never be reached in production? :)

lib/vsc/utils/fancylogger.py
boegel: No, it doesn't... We'll need a separate unit test for it (see test/fancylogger.py)

lib/vsc/utils/fancylogger.py
riccardomurri: Done in commit a13b6a6

lib/vsc/utils/fancylogger.py
riccardomurri: Done in commit 9a3f716

lib/vsc/utils/fancylogger.py
stdweird: how is the code below an enum instead of a namedtuple?

stdweird: also
```
a=namedtuple('ABC', ['x','y','z'])(*range(3))
```

riccardomurri: > how is the code below an enum instead of a namedtuple?

http://stackoverflow.com/a/1695250/459543

the (semantic) difference being that a namedtuple has independent fields which can be separately assigned, whereas an enum is just a bunch of constants

riccardomurri: Anyway, done in commit 5e9b507

boegel:
@riccardomurri are you planning to follow up on this, i.e. make the tests happy?

riccardomurri:
@boegel It's not on me -- the tests would be happy if you installed the updated version of vsc-base with colored logs support ;-)

stdweird:
@riccardomurri can you rebase on master and bump the version to 2.5.3?

riccardomurri:
@stdweird done.

riccardomurri:
I have no idea why Jenkins is complaining now. The relevant part of the console
output seems to be::

      File "/var/lib/jenkins/.local/lib/python2.7/unittest/loader.py", line 100, in loadTestsFromName
        parent, obj = obj, getattr(obj, part)
    AttributeError: 'module' object has no attribute 'fancylogger'

However I did not touch the test loader at all so I have no clue why this is
failing now.

JensTimmerman:
@riccardomurri this message is indeed a bit cryptic, it means the test/fancylogger.py file could not be imported:
```
>>> from test import fancylogger
  File "test/fancylogger.py", line 56, in <module>
    import coloredlogs
ImportError: No module named coloredlogs
```

riccardomurri:
Thanks @JensTimmerman !

Then the issue is: the `coloredlogs` dependency is only installed by `setup.py` when installing the optional component of the same name (e.g., `pip install -e vsc_base[coloredlogs]`).  So you would need to change the Jenkins script to perform the extended setup with optional components instead of the base one.

JensTimmerman:
@riccardomurri jenkins doesn't install the package, it just runs `python setup.py test`
if you put the extra requires also in `tests_require` it will install them.

riccardomurri:
Thanks again @JensTimmerman !  However, setting `tests_require` seems to break an unrelated test::

    ======================================================================
    FAIL: test_prospector (vsc.install.commontest.CommonTest)
    Run prospector.run.main, but apply white/blacklists to the results
      File "/var/lib/jenkins/.vsc/lib/python/vsc_install-0.10.13-py2.7.egg/vsc/install/commontest.py", line 177, in test_prospector
        self.assertTrue(False, 'prospector must be installed in jenkins environment')
    AssertionError: False is not true : prospector must be installed in jenkins environment

Is `prospector` a Python package?  Should I just add it to `tests_require`?

riccardomurri:
Adding `prospector` yields another seemingly unrelated test error::

    ======================================================================
    FAIL: test_timeout (test.run.TestRun)
      File "/data/jenkins/workspace/vsc-base-prbuilder/PYTHON/CPython-2.7/test/run.py", line 148, in test_timeout
        do_test(False)
      File "/data/jenkins/workspace/vsc-base-prbuilder/PYTHON/CPython-2.7/test/run.py", line 126, in do_test
    AssertionError: True is not false : main depth=2 pid (pids [28861, 28859, 28858]) is killed by timeout

What now?

JensTimmerman:
jenkins test this please

JensTimmerman:
@riccardomurri prospector should be added automatically by `vsc.install.shared_setup` but there seems to be a bug there, I'm looking into it.

JensTimmerman:
jenkins test this please

JensTimmerman:
jenkins test this please

JensTimmerman:
jenkins test this please

JensTimmerman:
fixed in vsc-install in https://github.com/hpcugent/vsc-install/pull/53 (already on jenkins)

JensTimmerman:
@riccardomurri tests work now, thank you very much for your patience.
however, is it possible to also set the background color when you do colored output?
since now it seems my INFO/WARNING/DEBUG labels are disappearing
screenshot from 2016-09-01 15-06-00
```
== 2016-09-01 15:05:14 templates.py:210 DEBUG name: versionprefix, config: 
== 2016-09-01 15:05:14 easyblock.py:2138 INFO Running method fetch_step part of step fetch
== 2016-09-01 15:05:14 easyblock.py:1408 WARNING Easyconfig does not specify an EasyBuild-version (key 'easybuild_version')! Assuming the latest version
== 2016-09-01 15:05:14 filetools.py:217 INFO Command eb found at /home/jens/.local/bin/eb
== 2016-09-01 15:05:14 tools.py:257 DEBUG Also considering installation prefix /home/jens/.local...
== 2016-09-01 15:05:14 tools.py:262 DEBUG Checking for easybuild/easyconfigs at /home/jens
```

riccardomurri:
Hi @JensTimmerman

I'm glad to know tests work now. Thanks for your quick reaction!

Apparently it's not possible to set the background, but you can change the foreground color of any logging level and field using environmental variables: http://coloredlogs.readthedocs.io/en/latest/#changing-the-colors-styles

(The default color scheme looks fine on black- or white- background terminals.)

JensTimmerman:
good enought for now, Thank you very much for this effort!

PR #226, opened 2016-03-26
Update setup.py

lowks:
Adding `description`


stdweird:
@lowks If you want to change the description, you should do so by changing the the description block in the `README.md`. It is extracted from the README during `sdist` via the `vsc-install` `shared_setup` magic/hack.

PR #227, opened 2016-05-03
fix tests and bugs for new vsc-install

JensTimmerman:
fix a tests that was broken in new vsc-install, + fix bad default arguments, and functions without engouht arguments

### DIFF
--- a/lib/vsc/utils/rest.py
+++ b/lib/vsc/utils/rest.py
@@ -104,7 +104,7 @@ class Client(object):
         elif token is not None:
             self.auth_header = '%s %s' % (token_type, token)
 
-    def get(self, url, headers={}, **params):
+    def get(self, url, headers=None, **params):
         """
         Do a http get request on the given url with given headers and parameters
         Parameters is a dictionary that will will be urlencoded
@@ -114,7 +114,7 @@ class Client(object):
         url += self.urlencode(params)
         return self.request(self.GET, url, None, headers)
 
-    def head(self, url, headers={}, **params):
+    def head(self, url, headers=None, **params):
         """
         Do a http head request on the given url with given headers and parameters
         Parameters is a dictionary that will will be urlencoded
@@ -124,7 +124,7 @@ class Client(object):
         url += self.urlencode(params)
         return self.request(self.HEAD, url, None, headers)
 
-    def delete(self, url, headers={}, **params):
+    def delete(self, url, headers=None, **params):
         """
         Do a http delete request on the given url with given headers and parameters
         Parameters is a dictionary that will will be urlencoded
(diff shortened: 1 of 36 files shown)


setup.py
```diff
@@ -36,10 +36,10 @@
 import vsc.install.shared_setup as shared_setup
 from vsc.install.shared_setup import ag, kh, jt, sdw
 
-VSC_INSTALL_REQ_VERSION = '0.9.19'
+VSC_INSTALL_REQ_VERSION = '0.10'
 
 PACKAGE = {
-    'version': '2.4.18',
+    'version': '2.5',
```
boegel: please stick to triple-digit versions? I hate it when packages go back & forth between two/three digit versions...

boegel: `2.5.0`, to keep it consistent with other versions like `2.4.18` and the upcoming `2.5.1`

'digits' is not exactly right, but you get the idea

JensTimmerman: so let's just enforce [semantic versioning](http://semver.org/) you mean? I agree, but let's do this for everything then, I'll also start commenting on version numbers in the future :-)

boegel: well, seems like you didn't for the last vsc-install, but there it was impossible to see...

I think we're pretty much sticking to semver otherwise?

stdweird: @JensTimmerman please keep the release number, with `.0` as the first release

setup.py
```diff
@@ -36,10 +36,10 @@
 import vsc.install.shared_setup as shared_setup
 from vsc.install.shared_setup import ag, kh, jt, sdw
 
-VSC_INSTALL_REQ_VERSION = '0.9.19'
+VSC_INSTALL_REQ_VERSION = '0.10'
```
stdweird: and make this `0.10.1`

lib/vsc/utils/rest.py
```diff
@@ -104,7 +104,7 @@ def __init__(self, url, username=None, password=None, token=None, token_type='To
         elif token is not None:
             self.auth_header = '%s %s' % (token_type, token)
 
-    def get(self, url, headers={}, **params):
+    def get(self, url, headers=None, **params):
```
boegel: you probably want to add this as well?

```python
if headers is None:
    headers = {}
```

JensTimmerman: these all pass this arg to self.request, and I've done that in self.request, so its ok.

lib/vsc/utils/rest.py
```diff
@@ -168,6 +168,8 @@ def patch(self, url, body=None, headers={}, **params):
         return self.request(self.PATCH, url, json.dumps(body), headers)
 
     def request(self, method, url, body, headers):
+        if headers is None:
```
boegel: missing docstring

PR #230, opened 2016-05-18
added test + fixes

JensTimmerman:


test/rest.py
```diff
@@ -68,9 +69,17 @@ def test_client(self):
         self.assertEqual(status, 200)
         self.assertEqual(body['merge_commit_sha'], u'fba3e13815f3d2a9dfbd2f89f1cf678dd58bb1f1')
 
-def suite():
-    """ returns all the testcases in this module """
-    return TestLoader().loadTestsFromTestCase(RestClientTest)
-
-if __name__ == '__main__':
-    main()
```
boegel: hmm, why not keep this? I find it useful to be able to run test modules in isolation?

test/rest.py
```diff
@@ -68,9 +69,17 @@ def test_client(self):
...
+        status, body = self.client.head()
+        self.assertEqual(status, 200)
+        try:
+            status, body = self.client.user.emails.post(body='jens.timmerman@ugent.be')
+            self.assertTrue(False, 'posting to unauthorized endpoint did not trhow a http error')
+        except HTTPError:
+            pass
+        try:
+            status, body = self.client.user.emails.delete(body='jens.timmerman@ugent.be')
+            self.assertTrue(False, 'deleting to unauthorized endpoint did not trhow a http error')
```
boegel: what about `patch` and `put`?

also: use a loop?

lib/vsc/utils/rest.py
```diff
@@ -124,15 +125,18 @@ def head(self, url, headers=None, **params):
...
         return self.request(self.HEAD, url, None, headers)
 
-    def delete(self, url, headers=None, **params):
+    def delete(self, url, headers=None, body=None, **params):
         """
-        Do a http delete request on the given url with given headers and parameters
+        Do a http delete request on the given url with given headers, body and parameters
         Parameters is a dictionary that will will be urlencoded
         """
         if self.append_slash:
```
stdweird: the `and not url.endswith('/')` is missing

lib/vsc/utils/rest.py
```diff
@@ -124,15 +125,18 @@ def head(self, url, headers=None, **params):
...
             url += '/'
         url += self.urlencode(params)
-        return self.request(self.DELETE, url, None, headers)
+        if headers is None:
```
stdweird: please stop copy pasting the same data over and over. use  a small function or even a decrator. (or move the same code to the `self.request`)

test/rest.py
stdweird: what does that even mean? you can use `python setup.py test` `-F`  and `-f` flags.

boegel: OK, fine.

lib/vsc/utils/rest.py
boegel: that's too late, since the `Content-Type` key is being updated below...

that was the bug that was introduced, various methods in `rest.py` are broken in current `master` because of this

stdweird: ? it's updated the same way, but for some reason you can't move this to `request` either?

boegel: ah, ok, you mean pass down the `application/json` to `self.request` which handles updating `Content-Type`

ok, works for me

boegel: I didn't notice that the values for `Content-Type` were all the same, so definitely +1 on @stdweird's suggestion to move that down into `request`

JensTimmerman:
OK, first of, this is supposed to fix errors not a rewrite, I'm not going to spend any more time on this, feel free to merge this or leave it as is.

since there's a lot of things that can be cleaned up here, there are a few ways of moving forward:
- merge this and come back later to clean everything up
- add a deprecated log call to it, pointing to upstream (which has become pretty active since we copy pasted this code) and merge this
- revert to the previous behavior and add a deprecated log to it, pointing to upstream

boegel:
Please don't abandon this, since you've broken the methods in `rest.py`, some of which are used in EB (and I'm sure in many other places).

JensTimmerman:
all 3 of the above options will fix the broken methods

boegel:
@JensTimmerman once you copy-paste code, you also take on the burden of maintaining it...

JensTimmerman:
the sticky fingers pattern is actually an anti pattern
http://c2.com/cgi/wiki?StickyFingers

boegel:
Yeah, well, screw patterns, let fix the bug you introduced, shall we?

I fixed the remark for you in https://github.com/JensTimmerman/vsc-base/pull/5, please merge that, and than this is good to go. Easy, huh?

boegel:
Thank you.

stdweird:
@JensTimmerman if upstream code now works/ is usable, feel free to look for an upgrade path. if can get rid ofmainiting code, i'm all in favour. 
wrt the fix, well, let's say we should just merge this and not comment and replacing copypaste with other copypaste

test/rest.py
stdweird: @JensTimmerman are you gong to address this? if not, merge this and replace it with upstrem in another PR

boegel:
to complete the `append_slash` cleanup: https://github.com/JensTimmerman/vsc-base/pull/6

test/rest.py
JensTimmerman: I have no good examples of what a patch and put actually should do at the moment.

PR #234, opened 2016-07-26
added Run async to file

Iepoev:


stdweird:
@Iepoev run async and run to file should not be combined (well, not like this), see e.g. the read_from_process of RunToFile.
you need to figure out if Popen with stdout a filehandle has bufferening (or is blocking).

Issue #237, opened 2016-09-05
undefined go_columns breaks generaloption in headless mode


boegel:
```
(Pdb) Traceback (most recent call last):
  File "/user/scratch/gent/vsc400/vsc40023/easybuild_easy_installed/lib/python2.7/site-packages/easybuild_framework-2.9.0.dev0-py2.6.egg/easybuild/tools/options.py", line 952, in parse_options
    with_include=with_include)
  File "/user/scratch/gent/vsc400/vsc40023/easybuild_easy_installed/lib/python2.7/site-packages/easybuild_framework-2.9.0.dev0-py2.6.egg/easybuild/tools/options.py", line 213, in __init__
    super(EasyBuildOptions, self).__init__(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/vsc/utils/generaloption.py", line 935, in __init__
    set_columns(kwargs.pop('go_columns', None))
  File "/usr/lib/python2.7/site-packages/vsc/utils/generaloption.py", line 69, in set_columns
    cols = int(os.popen('%s size 2>/dev/null' % stty).read().strip().split(' ')[1])
IndexError: list index out of range
```

cc @wpoely86

PR #238, opened 2016-09-05
also take into account possible IndexError in get_columns

boegel:
This fixes the bug reported in #237


itkovian:
If only there was a simple way to see what can be raised :(

PR #245, opened 2016-12-14
 have setLogLevel set the global log level instead of the fancylogger log level

JensTimmerman:
This will show a lot of logging, also for all dependencies etc.

### DIFF
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -722,7 +722,7 @@ def setLogLevel(level):
     """
     if isinstance(level, basestring):
         level = getLevelInt(level)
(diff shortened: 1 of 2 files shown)


JensTimmerman:
this still keeps fancylogger in it's namespace, but setLogLevel now sets the global loglevel for all new loggers.

stdweird:
@JensTimmerman  this pr does soemthing very different then #246 
please close one of them

JensTimmerman:
@stdweird which one?

stdweird:
@JensTimmerman if i can replace the logging rootlogger with fancylogger, i'd say both. i gave it some more tought, this is not the way to proceed; and if you want to change the root logger loglevel, use `logging`. but lets wait for what i come up with.

PR #246, opened 2016-12-14
let Generaloption set global log level

JensTimmerman:


stdweird:
i don't think this is the way to go. it's more useful that generaloption by default sets the root logger as a fancylogger. i'll make a PR soonish

boegel:
@stdweird is it soonish yet?

We're still using `logging.info` and `logging.debug` in `irun, but without making sure that `logging` uses a `FancyLogger`, that's not really usable to get decent log messages...

stdweird:
@boegel  is already merged https://github.com/hpcugent/vsc-base/pull/248

stdweird:
@JensTimmerman are these PRs still required?

boegel:
Ah, ok, I overlooked that because no new version was tagged, fixed that now, thanks @stdweird

PR #250, opened 2017-03-22
add support for providing custom log function in log.deprecated via log_callback

boegel:


lib/vsc/utils/fancylogger.py
```diff
@@ -345,6 +345,8 @@ def deprecated(self, msg, cur_ver, max_ver, depth=2, exception=None, *args, **kw
             self.raiseException("DEPRECATED (since v%s) functionality used: %s" % (max_ver, msg), exception=exception)
         else:
             deprecation_msg = "Deprecated functionality, will no longer work in v%s: %s" % (max_ver, msg)
+            if verbose:
+                sys.stderr.write(deprecation_msg)
```
stdweird: what's the point of this? too lazy to grep logfile? logging shouldn't mess with stderr directly. setup another handler that uses stderr and redirect all error and warnings to it if you really want this

boegel: deprecation warnings now only go to the log, which makes them practically invisible, you typically only dive into a log when something goes wrong

warnings should be printed when using deprecated functionality, like many other tools do too

if we don't want to do this in fancylogger, that's fine, then I'll construct some kind of wrapper around `log.deprecated` in EasyBuild

stdweird: if you don't care about other warnigns or errors is in your logfiles, why do you care about deprecation warnings?

boegel: Well, we don't use `log.error` in EasyBuild (we raise an exception instead), and `log.warning` is used for a bunch of things that may signal a problem but usually fix themselves later (e.g. not being able to resolve templates at an early stage)

I guess it boils down to semantics, deprecation warnings are another 'level' of warnings to me.

stdweird: then i propose that you pass another function instead of this variable, call that function on deprection, and have it default to `self.warning`

boegel:
Jenkins: test this please

lib/vsc/utils/fancylogger.py
```diff
@@ -345,7 +345,10 @@ def deprecated(self, msg, cur_ver, max_ver, depth=2, exception=None, *args, **kw
             self.raiseException("DEPRECATED (since v%s) functionality used: %s" % (max_ver, msg), exception=exception)
         else:
             deprecation_msg = "Deprecated functionality, will no longer work in v%s: %s" % (max_ver, msg)
-            self.warning(deprecation_msg)
+            if log_callback is None:
+                self.warning(deprecation_msg)
```
stdweird:
```python
if log_callback is None:
    log_callback = self.warning
```

PR #254, opened 2018-03-01
Singleton skip

stdweird:


test/patterns.py
```diff
@@ -0,0 +1,66 @@
...
+
+from vsc.utils.patterns import Singleton
+
+
+class ABC(object):
+    __metaclass__ = Singleton
+
+    def __init__(self, value):
+        """
+        Singleton metaclass with init that has args is probably the most stupid thing to have,
```
itkovian: I do not grok this comment. Why not be able to set stuff?

stdweird: check the tests, tell me that it makes sense

test/patterns.py
```diff
@@ -0,0 +1,66 @@
...
+
+        abc = ABC(1)
+        self.assertEqual(abc.value, 1, "init with value 1")
+
+        abc2 = ABC(2)
+        self.assertEqual(abc2.value, 1,
+                         "2nd init with value 2, gives value from (first) singleton")
+        self.assertEqual(id(abc2), id(abc), "2nd instance is identical to 1st")
+
+        abc3 = ABC(2, skip_singleton=True)
```
itkovian: Wait wut? :) So not a singleton after all.

test/sandbox/testpkg/__init__.py
```diff
@@ -1,4 +1,29 @@
...
+# vsc-base is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Library General Public License for more details.
+#
+# You should have received a copy of the GNU Library General Public License
+# along with vsc-base. If not, see <http://www.gnu.org/licenses/>.
+#
+#
 # Copyright 2016-2017 Ghent University
```
itkovian: Remove the below lines, they are double?

test/patterns.py
stdweird: yes, so we can transition away from it in eg `VSC()`

stdweird:
unittests fail, will have alook in a few days

test/patterns.py
JensTimmerman: @itkovian if 2 things need the singleton to be initialised in a different way, it can't and shouldn't be a singleton?

JensTimmerman:
this is a very ugly hack

PR #255, opened 2018-04-11
add get all groups code (HPC-6592)

JensTimmerman:
on top of cleanup


boegel:
@JensTimmerman So #256 should be merged first? Please indicate that in the description.

JensTimmerman:
This is the only way to reliably get all groups a users belongs to in python when using sssd as a backend. /etc/passwd don't contain all users, and the normal python functions don't return all users.
This implements a call to the same c function which  `groups <username> ` makes a call to, and wich does in fact returns the correct result. see also https://stackoverflow.com/questions/9323834/python-how-to-get-group-ids-of-one-username-like-id-gn

wdpypere:
@boegel think this needs to have a rereview.

boegel:
@wdpypere It's still unclear to me why this is needed, i.e. what problem it solves. I'm unaware of any issues where this is required.

Also: tests fail, there's little point in reviewing this in detail until the tests are fixed imho.

JensTimmerman:
this adds a python function to get all groups a user belongs too, in case this is ever needed.
Since python's grp doesn't talk with sssd, but tries /etc/passwd and others, which don't contain all users.
The tests fail because it's hard to test, it would seem the travis and jenkins system contain weird users, needs to be debugged indeed.

stdweird:
replced by #281

PR #256, opened 2018-04-13
Cleanup

JensTimmerman:


lib/vsc/utils/generaloption.py
```diff
@@ -683,6 +683,7 @@ def _is_enable_disable(x):
                 opt._long_opts = [x for x in opt._long_opts if not _is_enable_disable(x)]
         return fh
 
+    # pylint: disable=arguments-differ
```
boegel: @JensTimmerman Rather than ignoring it, can/should we fix it?

lib/vsc/utils/missing.py
```diff
@@ -257,12 +257,12 @@ def __init__(self, *args, **kwargs):
                 self.log.raiseException(msg, exception=KeyError)
 
         super(FrozenDictKnownKeys, self).__init__(tmpdict)
-
+    # pylint: disable=arguments-differ
```
boegel: same here, should we fix rather than ignore?

lib/vsc/utils/optcomplete.py
```diff
@@ -360,7 +361,7 @@ def print_usage_nousage (self, *args, **kwargs): # pylint: disable=unused-argume
     prev_interspersed = gparser.allow_interspersed_args  # save state to restore
     gparser.disable_interspersed_args()
 
-    cwords = os.environ.get('COMP_WORDS', '').split()
+    cwords = shlex.split(os.environ.get('COMP_WORDS', '').strip('() '))
```
boegel: @JensTimmerman Can you explain this? Seems like this is beyond cleanup, and an actual bug fix instead?

lib/vsc/utils/optcomplete.py
```diff
@@ -617,19 +618,19 @@ def gen_cmdline(cmd_list, partial, shebang=True):
     @param partial: the string to autocomplete (typically, partial is an element of the cmd_list)
     @param shebang: script has python shebang (if not, add sys.executable)
     """
-    cmdline = " ".join(cmd_list)
+    cmdline = ' '.join(['"%s"' % cmd for cmd in cmd_list])
```
boegel: I know why you did this, but this may look weird when we run into this again later; add a comment why you're injecting double quotes around the commands?

Also, what if the command includes double quotes already?

lib/vsc/utils/generaloption.py
JensTimmerman: no, this is by design, no way to fix this

lib/vsc/utils/optcomplete.py
JensTimmerman: yep, tests fail when a directory in the path has spaces in it, so have to do this to make jenkins happy. Should I create separate pr? all of this cleanup is just to make jenkins happy...

JensTimmerman:
@stdweird care to review?

lib/vsc/utils/optcomplete.py
boegel: Maybe mention that in a comment to clarify?

lib/vsc/utils/missing.py
```diff
@@ -40,8 +40,11 @@
 @author: Stijn De Weirdt (Ghent University)
 """
 import shlex
-import subprocess
 import time
+try:
+    from shlex import quote  # python 3.3
+except ImportError:
+    from pipes import quote  # python 2.7
```
JensTimmerman: actually already available in 2.5

JensTimmerman:
@stdweird review? This is needed for https://github.com/hpcugent/vsc-base/pull/255 which is needed before we disable ldap enumerate in sssd.conf

stdweird:
retest this please

stdweird:
@JensTimmerman is ok for me. but there's a test not running? and if this is merged, we need to push a new release to pypi, so i'll wait for it till tomorrow. ping me, and it'll try to merge it when you are ready to publish a new release

JensTimmerman:
@stdweird I accidentally pushed this branch to hpcugent fork instead of my fork, that's why jenkins is confused, but the test for this branch is working, so good to merge

stdweird:
@JensTimmerman commandline instruction give a merge conflict, feel free to do this yourself

boegel:
@JensTimmerman Can you sync your branch with current `master`, to re-trigger another test?

PR #261, opened 2018-05-23
Fix runqa empty output

itkovian:


boegel:
@itkovian Missing test? (and also needs a sync with `master`)

boegel:
Why was this merged without a test? :(

PR #263, opened 2018-05-25
specify that long description is in Markdown format

boegel:
Fix for problem with uploading to PyPI:

```
$ python setup.py sdist upload
...
INFO: running upload
Upload failed (400): The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.
error: Upload failed (400): The description failed to render in the default format of reStructuredText. See https://pypi.org/help/#description-content-type for more information.
```

It's a bit strange this hasn't popped up before; in the long run, we should probably fix this in `vsc-install` instead, but let's see if this fixes the issue first...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ _coloredlogs_pkgs = [
 ]
 
 PACKAGE = {
-    'version': '2.7.0',
+    'version': '2.7.1',
     'author': [sdw, jt, ag, kh],
     'maintainer': [sdw, jt, ag, kh],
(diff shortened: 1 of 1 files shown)


boegel:
OK, so that doesn't seem to make any difference... Ugh :(

boegel:
This explains it:

```
/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py:267: UserWarning: Unknown distribution option: 'long_description_content_type'
```

I'll try to fix the problems that pop up with `python setup.py check -r -s`

PR #264, opened 2018-05-25
fix syntax problems in README that block upload to PyPI

boegel:


.travis.yml
```diff
@@ -4,7 +4,12 @@ python:
   - "2.7"
 install:
   - easy_install vsc-install
+  # docutils to run "python setup.py check"
```
stdweird: should be moved to vsc-install.
just fix the issue in this PR, and open a PR or issue in vsc-install for a test

boegel: https://github.com/hpcugent/vsc-install/issues/87

PR #266, opened 2018-06-04
don't set executable when not using a shell

JensTimmerman:


lib/vsc/utils/fancylogger.py
```diff
@@ -740,9 +740,7 @@ def setLogLevel(level):
     """
     if isinstance(level, basestring):
         level = getLevelInt(level)
-
-    logger = getLogger(fname=False, clsname=False)
-
+    logger = logging.getLogger()
```
boegel: @JensTimmerman What's the point of this, in this PR? It even makes tests fail it seems...

setup.py
```diff
@@ -44,7 +44,7 @@
 ]
 
 PACKAGE = {
-    'version': '2.7.3',
+    'version': '2.7.5',
```
boegel: why not `.4`?

test/run.py
```diff
@@ -92,6 +92,15 @@ def test_noshell_glob(self):
         self.assertEqual(ec, 0)
         self.assertTrue(all(x in output.lower() for x in ['__init__.py', 'testmodule.py', 'testmodulebis.py']))
 
+    def test_noshell_executable(self):
+        ec, output = run("""%s -c 'print ("foo")'""" % sys.executable)
+        self.assertEqual(ec, 0)
+        self.assertTrue('foo' in output)
+        ec, output = run_simple([sys.executable, '-c', 'print ("foo")'])
```
boegel: `run_simple` expects a string, while `run` expects and array (you have them the other way around)?

boegel:
better/proper fix in #267...

test/run.py
JensTimmerman: no, run_simple also works with an array

JensTimmerman: and run also works with a string, but it's going to be very very escaped

PR #267, opened 2018-06-04
fix default value for SHELL in RunNoShell (should be None)

boegel:


lib/vsc/utils/run.py
```diff
@@ -448,6 +449,7 @@ def stop_tasks(self):
 
 class RunNoShell(Run):
     USE_SHELL = False
+    SHELL = None
```
boegel: This must be `None` by default since it doesn't make sense to specify a value for `executable` in the `subprocess.Popen` call when we instruct it to not use a shell.

With the current default, shell commands fail with weird errors like this:

```
$ python -c "from vsc.utils.run import run; ec, out = run(['echo', 'hello']); print ec; print out"
2018-06-04 17:46:57,797 ERROR      <string>.RunNoShell MainThread  _post_exitcode: problem occured with cmd ['echo', 'hello']: (shellcmd ['echo', 'hello']) output echo: hello: No such file or directory

127
echo: hello: No such file or directory
```

on top of this branch, this now works fine:

```
$ python -c "from vsc.utils.run import run; ec, out = run(['echo', 'hello']); print ec; print out"
0
hello
```

boegel: To explain the error: if the `executable` argument (taken from `SHELL`) for `subprocess.Popen` is not `None`, then the first element of the command is "consumed", which leads to "`hello`" being executed rather than "`echo hello`", which explains the `No such file or directory` error...

PR #270, opened 2018-06-06
add CmdList wrapper for list type as helper for constructing commands

boegel:
`mympirun` is currently broken because of the naive switch from `run_simple` to `run` in https://github.com/hpcugent/vsc-mympirun/pull/144 .

When using `run` with a list to specify the command, *none* of the items in the list should be a combination of separate arguments, otherwise they will be treated as an argument as a whole, leading to errors like `unrecognized argument machinefile  /user/home/gent/vsc400/vsc40023/.mympirun_bfxlm3/4473084.master15.delcatty.gent.vsc_20180606_151207/nodes` (where `machinefile` and the `.../nodes` path are supposed to be passed as separate arguments rather than as a string combining both).

In addition, all elements in the command list *must* be strings. If not, it leads to errors like:

```
  File "/user/scratchphanpy/gent/gvo000/gvo00002/vsc40023/easybuild_REGTEST/CO7/sandybridge/software/vsc-mympirun/4.1.4dev/lib/python2.7/site-packages/vsc_mympirun-4.1.4-py2.7.egg/EGG-INFO/scripts/mympirun", line 124, in main
    instance.main()
  File "/user/home/gent/vsc400/vsc40023/eb_phanpyscratch/CO7/sandybridge/software/vsc-mympirun/4.1.4dev/lib/python2.7/site-packages/vsc_mympirun-4.1.4-py2.7.egg/vsc/mympirun/mpi/mpi.py", line 450, in main
    exitcode, _ = run_mpirun_cmd(self.mpirun_cmd, **run_kwargs)
  File "build/bdist.linux-x86_64/egg/vsc/utils/run.py", line 151, in run
    return r._run()
  File "build/bdist.linux-x86_64/egg/vsc/utils/run.py", line 255, in _run
    self._run_pre()
  File "build/bdist.linux-x86_64/egg/vsc/utils/run.py", line 273, in _run_pre
    self._init_process()
  File "build/bdist.linux-x86_64/egg/vsc/utils/run.py", line 367, in _init_process
    self._process = self._process_module.Popen(self._shellcmd, **self._popen_named_args)
  File "/user/home/gent/vsc400/vsc40023/eb_phanpyscratch/CO7/sandybridge/software/Python/2.7.14-intel-2018a/lib/python2.7/subprocess.py", line 390, in __init__
    errread, errwrite)
  File "/user/home/gent/vsc400/vsc40023/eb_phanpyscratch/CO7/sandybridge/software/Python/2.7.14-intel-2018a/lib/python2.7/subprocess.py", line 1025, in _execute_child
    raise child_exception
TypeError: execv() arg 2 must contain only strings
```

The `CmdList` class should help with avoiding such situations...

### DIFF
--- a/lib/vsc/utils/run.py
+++ b/lib/vsc/utils/run.py
@@ -86,6 +86,48 @@ BASH = '/bin/bash'
 SHELL = BASH
 
 
+class CmdList(list):
+    """Wrapper for 'list' type to be used for constructing a list of options & arguments for a command."""
+
+    def __init__(self, cmd=None):
+        """
+        Create CmdList instance to construct command
+
+        :param cmd: actual command to run (first item in list)
+        """
+        super(CmdList, self).__init__()
+        if cmd is not None:
+            if isinstance(cmd, list):
+                super(CmdList, self).extend(cmd)
+            else:
+                super(CmdList, self).append(cmd)
+
+    def add_opts_args(self, items, tmpl_vals=None, allow_spaces=True):
+        """
+        Add options/arguments to command
+
+        :param item: option/argument to add to command
+        :param tmpl_vals: template values for item
+        """
+        if isinstance(items, basestring):
+            items = [items]
+
+        for item in items:
+            if tmpl_vals:
+                item = item % tmpl_vals
+
+            if not allow_spaces and ' ' in item:
+                raise ValueError("Found one or more spaces in item '%s' being added to command %s" % (item, self))
+
+            super(CmdList, self).append(item)
+
+    def append(self, *args, **kwargs):
+        raise NotImplementedError("Use add_opts_args rather than append")
+
+    def extend(self, *args, **kwargs):
+        raise NotImplementedError("Use add_opts_args rather than extend")
+
+
 class DummyFunction(object):
     def __getattr__(self, name):
         def dummy(*args, **kwargs):  # pylint: disable=unused-argument
--- a/test/run.py
+++ b/test/run.py
@@ -38,7 +38,7 @@ import time
 import shutil
 from unittest import TestLoader, main
 
-from vsc.utils.run import run, run_simple, run_asyncloop, run_timeout, RunQA, RunTimeout
+from vsc.utils.run import CmdList, run, run_simple, run_asyncloop, run_timeout, RunQA, RunTimeout
 from vsc.utils.run import RUNRUN_TIMEOUT_OUTPUT, RUNRUN_TIMEOUT_EXITCODE, RUNRUN_QA_MAX_MISS_EXITCODE
 from vsc.install.testing import TestCase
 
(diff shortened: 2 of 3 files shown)


lib/vsc/utils/run.py
```diff
@@ -86,6 +86,48 @@
...
+        :param cmd: actual command to run (first item in list)
+        """
+        super(CmdList, self).__init__()
+        if cmd is not None:
+            if isinstance(cmd, list):
+                super(CmdList, self).extend(cmd)
+            else:
+                super(CmdList, self).append(cmd)
+
+    def add_opts_args(self, items, tmpl_vals=None, allow_spaces=True):
```
JensTimmerman: why not make this append?

JensTimmerman:
This doesn't make a lot of sense to me? You're not really solving anything?

just turning
```
run([a,b,c])
```
into
```
d = CmdList()
d. add_opts_args(a)
d. add_opts_args(b)
d. add_opts_args(c)
run(d)
```
which doesn't look all that better to me...

lib/vsc/utils/run.py
```diff
@@ -86,6 +86,48 @@
...
+        """
+        Add options/arguments to command
+
+        :param item: option/argument to add to command
+        :param tmpl_vals: template values for item
+        """
+        if isinstance(items, basestring):
```
JensTimmerman: will break the python3 porting effort

lib/vsc/utils/run.py
```diff
@@ -86,6 +86,48 @@
...
+    """Wrapper for 'list' type to be used for constructing a list of options & arguments for a command."""
+
+    def __init__(self, cmd=None):
+        """
+        Create CmdList instance to construct command
+
+        :param cmd: actual command to run (first item in list)
+        """
+        super(CmdList, self).__init__()
+        if cmd is not None:
```
JensTimmerman: initial command is not sanity checked? can contain spaces, can be a non string thing?

lib/vsc/utils/run.py
boegel: i) to capture cases where `.append` and `.extend` is being used
ii) `add_opts_args` is a joint alternative for `append`/`extend`

lib/vsc/utils/run.py
boegel: We'll fix that when we actually get to porting `run.py` to Python 3 (we don't have a replacement for this yet that works with both Py2 & Py3).

lib/vsc/utils/run.py
boegel: Good point, I'll fix that by using `add_opts_args` here too.

lib/vsc/utils/run.py
```diff
@@ -86,6 +86,48 @@
...
+        """
+        Add options/arguments to command
+
+        :param item: option/argument to add to command
+        :param tmpl_vals: template values for item
+        """
+        if isinstance(items, basestring):
+            items = [items]
+
+        for item in items:
```
JensTimmerman: what if items is not itterable?

JensTimmerman:
If you're going to use this, I would add an append method that adds one argument, and extend which takes an iterable (that's not a basestring) and adds all these. Don't do 2 things in 1 function to that in that function have have special cases.
That way it's also more pythonic.

lib/vsc/utils/run.py
boegel: I wanted to avoid confusion with `set.add`, but fine, will rename to `.add`

boegel:
@JensTimmerman I disagree on separate methods for `append`/`extend`, can only lead to mistakes imho (and more code), I don't see the point

@stdweird W.r.t. `CmdList().run` support: which run should it use then? Also, can be done in a later PR, this is blocking https://github.com/hpcugent/vsc-mympirun/pull/145

lib/vsc/utils/missing.py
```diff
@@ -55,6 +55,11 @@
 _log = fancylogger.getLogger('vsc.utils.missing')
 
 
+def is_string(item):
```
JensTimmerman: should be in missing.py

test/run.py
```diff
@@ -284,3 +284,74 @@ def test_qa_list_of_answers(self):
...
+        cmd = CmdList('echo', "hello world", 'test')
+        self.assertEqual(cmd, ['echo', "hello world", 'test'])
+
+        # can also create via template
+        self.assertEqual(CmdList('%(one)s', tmpl_vals={'one': 1}), ['1'])
+
+        # adding non-string items yields an error
+        self.assertErrorRegex(ValueError, "Non-string item", cmd.add, 1)
+        self.assertErrorRegex(ValueError, "Non-string item", cmd.add, None)
+        self.assertErrorRegex(ValueError, "Non-string item", cmd.add, ['foo', None])
```
JensTimmerman: add test for adding list of lists, maybe have it show a clear error like: you can't add a list of lists (something an extend command might be able to do)

boegel: test case added
a list is not a string, so error is clear imho

PR #272, opened 2018-06-08
check whether /dev/log exists before logging to it in logToDevLog

boegel:
This mainly facilitates testing on systems where there's no `/dev/log` present.

Without this, `logToDevLog` fails in mysterious ways, without a trace of what the actual problem is (missing `/dev/log`):

```
python -c "from vsc.utils.fancylogger import logToDevLog; logToDevLog()"
  File "<string>", line 1, in <module>
  File "/Volumes/work/vsc-base/lib/vsc/utils/fancylogger.py", line 733, in logToDevLog
    syslogoptions, 'logtodevlog', enable=enable, name=name, handler=handler)
  File "/Volumes/work/vsc-base/lib/vsc/utils/fancylogger.py", line 648, in _logToSomething
    handler = handlerclass(**handleropts)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/handlers.py", line 762, in __init__
    self._connect_unixsocket(address)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/logging/handlers.py", line 790, in _connect_unixsocket
    self.socket.connect(address)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/socket.py", line 228, in meth
    return getattr(self._sock,name)(*args)
socket.error: [Errno 2] No such file or directory
```

If `/dev/log` is not there, logging continues just fine to the default syslog location.

### DIFF
--- a/lib/vsc/utils/fancylogger.py
+++ b/lib/vsc/utils/fancylogger.py
@@ -724,11 +724,13 @@ def _getSysLogFacility(name=None):
 
 def logToDevLog(enable=True, name=None, handler=None):
     """Log to syslog through /dev/log"""
+
+    syslogoptions = {'facility': _getSysLogFacility()}
+
     devlog = '/dev/log'
-    syslogoptions = {
-        'address': devlog,
-        'facility': _getSysLogFacility()
-    }
+    if os.path.exists(devlog):
+        syslogoptions['address'] = devlog
+
     return _logToSomething(logging.handlers.SysLogHandler,
                            syslogoptions, 'logtodevlog', enable=enable, name=name, handler=handler)
 
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ _coloredlogs_pkgs = [
 ]
 
 PACKAGE = {
-    'version': '2.8.1',
+    'version': '2.8.2',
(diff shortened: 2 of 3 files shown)


lib/vsc/utils/fancylogger.py
```diff
@@ -724,11 +724,13 @@ def _getSysLogFacility(name=None):
 
 def logToDevLog(enable=True, name=None, handler=None):
     """Log to syslog through /dev/log"""
```
stdweird: so this works without /dev/log? and does it work without th eoption when /dev/log is present?

boegel: Yes, it now works without `/dev/log`. It'll just log to the default syslog location on the system when `/dev/log` (e.g. `/var/log/system/log` on my laptop).

Not sure I get your second question, which option do you mean?

stdweird: can you try on one of our hosts? i'm quite sure this is not always possible.
i guess your laptop runs some syslog service that listen on some default port and accepts anything?

wdpypere:
@boegel conflicts, plus can you test on our systems as @stdweird requested?

PR #273, opened 2018-06-11
Test runqa

itkovian:


itkovian:
Not sure if this suffices.

boegel:
@itkovian Doesn't seem to be, the test passes as well on an older version of `vsc-base` that doesn't include the fix from #261...

stdweird:
i'm not sure how easy this will be to test. the issue it fixes was due to a timing issue, so that might be too hard. the only thing you can do is count function/methods calls in `run`, there should be only one hit for that question, that is what is fixed here.

itkovian:
The issue, iirc, was that after the QA we got an empty line and then an exit of the QA'ing process.

stdweird:
after the 1st reply, the process continued for a short fraction; then qa loop had another iteration and a hit (due to parsing almost the same, already replied-to output), and by the time some other code in the qa loop was executed, the process was stopped, and then the rest of the qa code tried to answer a non-running process and crashed.
there's no point in trying to repeat this in a test, the fundamental issue was that qa tried to answer something that it had already replied to (the output was changed by the extra whitespace, but that was not the main issue).

wdpypere:
tests fail

itkovian:
@stdweird So I remove the test? :-p

stdweird:
@itkovian i would just close the PR for now

PR #274, opened 2018-06-15
logging: add formatting to each line of a multiline log string

smoors:


lib/vsc/utils/fancylogger.py
```diff
@@ -631,7 +659,7 @@ def _logToSomething(handlerclass, handleropts, loggeroption,
     logger = getLogger(name, fname=False, clsname=False)
 
     if formatterclass is None:
-        formatterclass = logging.Formatter
+        formatterclass = MultilineFormatter
```
stdweird: this cannot become the default logger. the big problem with multiline splitting is that you cannot tell what belongs together.
i'm not sure what problem you want to solve, but this is not a general solution.

smoors: it's meant for easier grepping of easybuild logs.
I was not sure if it should go here or into easybuild, or whether it should be behind a pref.
what do you suggest?

stdweird: you can do what was done for the colourer: add the code here and make it default in EB
to be claer: i don't mind that the code is in here, but i do mind it becomes the default

stdweird: but i'm not sure why this will become easier. also, i thought EB was using a streamlogger, so i'm not so sure what that will do to it.

smoors: well, it works for most log string in EB so I think I'm doing some things right :)
not sure why it doesn't work for all though.
if every line in the logs is formatted, you could for example grep for the INFO lines, leaving all the DEBUG lines out.

smoors: maybe @boegel can explain?

boegel: I agree that it should somehow be clear that a log line is a continuation of a previous one... Maybe by inserting some kind of line header like `>>>` or `(cont.)>` before the message?

stdweird: currently, the lack of any prefix indicates continuation.
choose for the multilineformatter whatever format you want, as long as it isn't the default

smoors: but are you sure that output from different log commands (eg. when multiple processes try to write to the log file at the same time) will not intersperse?

stdweird: ofcourse not, but neither does your multiline fomratter. (but again, add the code here, enable it in EB)

Issue #268, opened 2018-06-04
tests for rest.py fail in Travis because of API rate limiting


boegel:
```
ERROR: Test all request methods
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/hpcugent/vsc-base/test/rest.py", line 74, in test_request_methods
    status, body = self.client.head()
  File "lib/vsc/utils/rest.py", line 128, in head
    return self.request(self.HEAD, url, None, headers)
  File "lib/vsc/utils/rest.py", line 175, in request
    conn = self.get_connection(method, url, body, headers)
  File "lib/vsc/utils/rest.py", line 206, in get_connection
    connection = self.opener.open(request)
  File "/opt/python/2.6.9/lib/python2.6/urllib2.py", line 397, in open
    response = meth(req, response)
  File "/opt/python/2.6.9/lib/python2.6/urllib2.py", line 510, in http_response
    'http', request, response, code, msg, hdrs)
  File "/opt/python/2.6.9/lib/python2.6/urllib2.py", line 435, in error
    return self._call_chain(*args)
  File "/opt/python/2.6.9/lib/python2.6/urllib2.py", line 369, in _call_chain
    result = func(*args)
  File "/opt/python/2.6.9/lib/python2.6/urllib2.py", line 518, in http_error_default
    raise HTTPError(req.get_full_url(), code, msg, hdrs, fp)
HTTPError: HTTP Error 403: Forbidden
```

We can fix this by using a GitHub token in Travis, and skip the tests that require a token when the token is missing (token will never be available when testing PRs because of security reasons).

PR #276, opened 2019-02-07
actually return headers when http method is HEAD

migueldiascosta:
the message body when using the http HEAD method is empty, this change allows the head() method to actually return the headers (https://github.com/easybuilders/easybuild-framework/pull/2553#issuecomment-458456195)

### DIFF
--- a/bin/logdaemon.py
+++ b/bin/logdaemon.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #
-# Copyright 2011-2018 Ghent University
+# Copyright 2011-2019 Ghent University
 #
 # This file is part of vsc-base,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
--- a/lib/vsc/__init__.py
+++ b/lib/vsc/__init__.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2015-2018 Ghent University
+# Copyright 2015-2019 Ghent University
 #
 # This file is part of vsc-base,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
--- a/lib/vsc/utils/__init__.py
+++ b/lib/vsc/utils/__init__.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2015-2018 Ghent University
+# Copyright 2015-2019 Ghent University
 #
 # This file is part of vsc-base,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
--- a/lib/vsc/utils/affinity.py
+++ b/lib/vsc/utils/affinity.py
@@ -1,5 +1,5 @@
 #
-# Copyright 2012-2018 Ghent University
+# Copyright 2012-2019 Ghent University
 #
 # This file is part of vsc-base,
 # originally created by the HPC team of Ghent University (http://ugent.be/hpc/en),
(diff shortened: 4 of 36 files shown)


stdweird:
@migueldiascosta can you bump the version in setup.py ?
@boegel i guess this needs a backport to EB too?

migueldiascosta:
@stdweird done. I can submit the same change to the `4.x` branch in the EB framework

stdweird:
@migueldiascosta thanks
wrt the failing tests, i can fix the headers errors (or you can too, see "fix license headers" in https://github.com/hpcugent/vsc-install/#fix-failing-tests)
but can you have a look at the other failing test (as it looks related to the code you modified)?

migueldiascosta:
@stdweird the `test_request_methods` error does seem related to my change, since that starts with a call to head() (but it only checks the status, which comes from `get_connection`, before my change)

Also, `test_client` is also returning a 403, and that's a GET method, also failing at `get_connection` (and my change should only affect the HEAD method)

I may be missing something but isn't this github rate limiting or something related? If not, I'll come back for a better look

boegel:
Could definitely be related to rate limiting, we haven't set up a GitHub token for the tests in Travis like we have for EasyBuild (but we'd also need to disable tests for PRs that require a token, since tokens are not available for testing PRs in Travis, to avoid that they leak on automatic testing of PRs).

migueldiascosta:
@boegel @stdweird I think the rollback confirms that the problem is not with my change

also, running the tests manually with `python setup.py test -F rest` works for me, with and without the change, can you also test manually?

If that's the case, I can bring back my change, update the test to also check thad `head()` does return the headers and skip the tests if there is no token available, as in EB?

boegel:
@migueldiascosta Can you please re-instate the fix (by just removing the rollback commit + `git push --force`)?

The problem is indeed that we don't have a token in place for GitHub, see also https://github.com/hpcugent/vsc-base/issues/268.

It's not trivial to put that in place though, and it won't fix the problem for PRs anyway (unless we skip tests that fail because of rate limiting).

We can work around it temporarily by retriggering the tests in Travis that broke because of rate limiting, for now (as we've done in other recent PRs).

migueldiascosta:
@boegel done

what about also testing if head() is indeed returning the headers? something like

```diff
         """Test all request methods"""
         status, body = self.client.head()
         self.assertEqual(status, 200)
+        self.assertTrue(body)
+        self.assertIn('X-GitHub-Media-Type', body)
         try:
             status, body = self.client.user.emails.post(body='jens.timmerman@ugent.be')
             self.assertTrue(False, 'posting to unauthorized endpoint did not trhow a http error')
```

boegel:
@migueldiascosta Enhancing the tests to cover the change you made makes a lot of sense, so please do!

Tests are OK now after re-triggering them, which shows the only problem is the GitHub rate limiting...

boegel:
Green light now from Travis after a bit of patience & re-triggering tests, thanks @migueldiascosta!

stdweird:
@migueldiascosta thanks
wrt the failing tests, i can fix the headers errors (or you can too, see "fix license headers" in https://github.com/hpcugent/vsc-install/#fix-failing-tests)
but can you have a look at the other failing test (as it looks related to the code you modified)?

migueldiascosta:
@boegel shall I submit a PR to the `4.x` branch of EB?

boegel:
@migueldiascosta Please do!

PR #277, opened 2019-03-08
take into account slightly different output with bash 4.x in test_noshell_glob

boegel:
In bash 4.x, the error message looks like:

```
cannot access 'test/sandbox/testpkg/*': No such file or directory
```

while in bash 3.x it looks like:

```
test/sandbox/testpkg/*: No such file or directory
```

(i.e. no single quotes around it)

No need for a version bump imho, since this only touches the tests...


hajgato:
lgtm

PR #278, opened 2019-03-08
skip tests that talk to GitHub API when no GitHub token is available

boegel:
I've added a GitHub token in Travis using the `$VSC_GITHUB_TOKEN` environment variable, but this is not made available for pull requests (since then anyone could steal the token by just printing it via a PR).

Hence, we should skip the `rest` tests when no GitHub token is available, to avoid that the tests fail due to GitHub rate limiting...

(only changes in tests, so no need to bump the version imho)

### DIFF
--- a/test/rest.py
+++ b/test/rest.py
@@ -60,6 +60,10 @@ class RestClientTest(TestCase):
 
     def test_client(self):
         """Do a test api call"""
+        if GITHUB_TOKEN is None:
+            print("Skipping test_client, since no GitHub token is available")
+            return
+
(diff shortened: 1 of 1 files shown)


test/rest.py
```diff
@@ -60,6 +60,10 @@ def setUp(self):
 
     def test_client(self):
         """Do a test api call"""
+        if GITHUB_TOKEN is None:
```
stdweird: next time, please use the skip decorators (and ffs, don't print, but log)

PR #280, opened 2019-04-01
make prospector checks for Python 3 pass by improving compatibility with Python 3

boegel:
This is probably easiest to review commit by commit:

* changes in 7dd8fd0 are just for `print` statements, and are trivial; I should mention that I purposely did not use the oneliner mentioned in https://github.com/hpcugent/vsc-install#fixing-print-statement, because I noticed it incorrectly reformats `print foo, bar` to `print(foo, bar)` (the latter prints a tuple rather than two values separated by a space in Python 2...)

* changes in 6c3b175 are less trivial, and deserve a closer look

* changes in c7210f7 are just to fix use of `__metaclass__` in `Wrapper`; the fix is basically the same as in https://github.com/hpcugent/vsc-config/pull/84, but the inner class needs to be fleshed out first...

cc @stdweird @hajgato


lib/vsc/utils/daemon.py
```diff
@@ -123,7 +124,7 @@ def stop(self):
                 if os.path.exists(self.pidfile):
                     os.remove(self.pidfile)
             else:
-                print str(err)
+                print(str(err))
```
stdweird: logging.error

lib/vsc/utils/dateandtime.py
```diff
@@ -138,7 +138,7 @@ def number(self, otherdate):
     def get_other(self, shift=-1):
         """Return month that is shifted shift months: negative integer is in past, positive is in future"""
         new = self.date.year * 12 + self.date.month - 1 + shift
-        return self.__class__(date(new // 12, new % 12 + 1, 01))
+        return self.__class__(date(new // 12, new % 12 + 1, 0o1))
```
stdweird: you mean `1`

lib/vsc/utils/fancylogger.py
```diff
@@ -326,7 +327,10 @@ def raiseException(self, message, exception=None, catch=False):
             exception = self.RAISE_EXCEPTION_CLASS
 
         self.RAISE_EXCEPTION_LOG_METHOD(fullmessage)
-        raise exception, message, tb
+
+        err = exception(message)
```
stdweird: why is this not `exception(message, tb)`?

lib/vsc/utils/optcomplete.py
```diff
@@ -91,7 +91,7 @@
 #      An array variable from which Bash reads the possible completions
```
stdweird: you can remove this module

lib/vsc/utils/fancylogger.py
boegel: I based this change on https://github.com/hpcugent/vsc-install#old-raise-syntax, not sure if what you propose is compatible with both Python 2 & 3... @hajgato?

lib/vsc/utils/optcomplete.py
boegel: After our discussion this morning, keep it after all, or not?

lib/vsc/utils/fancylogger.py
hajgato: @stdweird  I have followed the recommendations on this page: https://portingguide.readthedocs.io/en/latest/exceptions.html#the-new-raise-syntax I do believe that your proposed syntax would not work in `Python 3`.

boegel: I discussed this with @hajgato, `exception(message, tb)` is most likely not compatible with Python 3.

On the other hand, fiddling with `err.__traceback__` does not work in Python 2.6 (and it probably explains why the tests fail in Travis for Python 2.6 while they pass for Python 2.7).

@stdweird Do we still care about Python 2.6 support? If not, I'll just drop the Python 2.6 config in Travis...

hajgato: https://legacy.python.org/dev/peps/pep-3109/

stdweird: @boegel yes, please drop 2.6

lib/vsc/utils/optcomplete.py
stdweird: remove it

boegel: It's imported from `generaloption`, not so trivial to flesh it out.

If you want it gone, please open an issue, won't do it in this PR.

PR #281, opened 2019-05-21
Add members in group lookup

stdweird:
Replaces #255


lib/vsc/utils/generaloption.py
```diff
@@ -564,30 +564,30 @@ def set_description_docstring(self):
         except (IndexError, ValueError, AttributeError):
             self.log.debug("set_description_docstring: no docstring found in latest stack globals")
             docstr = None
+            return None
```
boegel: not a big fan of inline `return` statements, but fine...

lib/vsc/utils/groups.py
```diff
@@ -0,0 +1,72 @@
...
+    grouplist = (c_uint * ngroups)()
+    ngrouplist = c_int32(ngroups)
+
+    if isinstance(user, basestring):
+        user = pwd.getpwnam(user)
+    else:
+        user = pwd.getpwuid(user)
+
+    ct = getgrouplist(user.pw_name, user.pw_gid, byref(grouplist), byref(ngrouplist))
+    # if a max of 50 groups was not enough, try again with exact given nr
```
boegel: Why not go with actual value from the start? What's the downside of doing so, that's not clear to me...
It would avoid quite a bit of duplicate code, no?

lib/vsc/utils/groups.py
```diff
@@ -0,0 +1,72 @@
...
+    if ct < 0:
+        getgrouplist.argtypes = [c_char_p, c_uint, POINTER(c_uint *int(ngrouplist.value)), POINTER(c_int32)]
+        grouplist = (c_uint * int(ngrouplist.value))()
+        ct = getgrouplist(user.pw_name, user.pw_gid, byref(grouplist), byref(ngrouplist))
+
+    if ct < 0:
+        raise Exception("Could not find groups for %s" % user)
```
boegel: Can the `ct` value be useful to tell what went wrong? If so, include it in error message?

stdweird: added

lib/vsc/utils/groups.py
stdweird: how would you know the actual value, it's what we are going to lookup

lib/vsc/utils/generaloption.py
stdweird: not a fan of pointless indents

stdweird:
@boegel version bumped

lib/vsc/utils/groups.py
boegel: OK, maybe it's worth clarifying that the first call to `getgrouplist` is required to be able to check `ngrouplist.value`, since that's not immediately clear.

How safe is the `50` initial value? Why `50`, how likely is it that it's not enough, and what goes wrong if it's not?

stdweird: 50 is just a guess, not too big, not too small. it's very unlikely that someone is in more than 50 groups, that's what's important. 
for how getgrouplist works, see http://man7.org/linux/man-pages/man3/getgrouplist.3.html

boegel: So if the initial guess of 50 is wrong, no harm done since there's a fallback to actually asking the number and then getting all the needed info, right?

If so, yeah, this is OK.

50 isn't terribly unlikely btw, it seems like my VSC account is relatively close (34 groups in accountpage + VO + a few more on the system probably)...

stdweird:
@boegel should ok now

test/groups.py
```diff
@@ -0,0 +1,61 @@
...
+        """Test getgrouplist"""
+        for user in pwd.getpwall():
+            gidgroups = sorted(getgrouplist(user.pw_uid))
+            namegroups = sorted(getgrouplist(user.pw_name))
+
+            # get named groups from id
+            #   grp.getgrall is wrong, e.g. for root user on F30,
+            #   gr_mem for root group is empty (as is entry in /etc/group),
+            #   while id returns 1 group
+            #groups = [g.gr_name for g in grp.getgrall() if user.pw_name in g.gr_mem]
```
JensTimmerman: Aha, that's what is going wrong, any idea why this is?

lib/vsc/utils/groups.py
JensTimmerman: @boegel, yeah, getgrouplist is weird, needing a number as input, failing if that number is wrong, but in the process telling you what the number should have been...

test/groups.py
stdweird: i think `grp` really only queries `/etc/group`, and even on F30 on my laptop sssd is running and handled first in nssswitch; but i don't see what sssd config default is repsonsible the behaviour

Issue #282, opened 2019-07-05
import of fancylogger module fails with Python 3.7


boegel:
For some reason I thought we had already ported `vsc-base` to Python 3, but it seems like I'm mis-remembering that... Some progress was made in #280, but it's definitely not complete.

```
PYTHONPATH=$HOME/work/vsc-base/lib python3 -c 'import vsc.utils.fancylogger'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/kehoste/work/vsc-base/lib/vsc/utils/fancylogger.py", line 192, in <module>
    logging._levelNames['EXCEPTION'] = logging.ERROR
AttributeError: module 'logging' has no attribute '_levelNames'
```

stdweird:
vsc-install and vsc-base are not supported on py3 atm

boegel:
@stdweird Yeah, I forgot about that after ingesting what we need in EasyBuild and porting stuff there. ;)

PR #283, opened 2019-07-05
fix import of fancylogger with Python 3

boegel:
Baby step towards supporting Python 3, I apparently had this sitting in a branch but never opened a PR for it (I knew #282 looked familiar).

More changes are needed to actually run the `fancylogger` tests, that would be the next goal I guess...


test/fancylogger.py
```diff
@@ -412,7 +418,7 @@ def somefunction(self):
         handler = fancylogger.logToScreen()
         logger = fancylogger.getLogger(fname=False, clsname=False)
         logger.warn("blabla")
-        print stringfile.getvalue()
+        print(stringfile.getvalue())
```
stdweird: should import from future ...

.travis.yml
```diff
@@ -1,9 +1,25 @@
...
 script:
   # 'hpcugent' is a mandatory remote for setup.py to work (because of vsc-install)
   - git remote add hpcugent https://github.com/hpcugent/vsc-base.git
-  - python setup.py test
+  # only run part of the tests with Python 3 for now (until whole code base is Py3 compatible)
+  - if [ "$(echo $TRAVIS_PYTHON_VERSION | cut -f1 -d.)" == '3' ]; then
+        set -e;
+        export PYTHONPATH="$PWD/lib:$PYTHONPATH";
+        python -c "import vsc.utils.fancylogger";
+        python -c "import test.fancylogger";
```
stdweird: this does not actually test anything, so why bother?

boegel: It saves some of the problems, makes the PR short and easy to review, etc., the tests pass as is, etc.

Baby steps.

boegel: To clarify:

* this checks whether `vsc/utils/fancylogger.py` no longer has `SyntaxError`s; same for `test/fancylogger.py`
* can't run only the `fancylogger` subsuite yet, more changes are needed for that due to `SyntaxError` in `test/missing.py`

If you prefer, I can fix the import `test/missing.py` too, and see if that's enough to run the `fancylogger` tests with Python 3 (but it's going to blow up the PR, why not merge it first since it's easy to review right now).

test/fancylogger.py
boegel: `import` is there now (but this change is still needed, `print 'foo'` never works in Python 3, even with the `import`

boegel:
Closing this in favor of #287

PR #285, opened 2019-11-22
run tests with both Python 2.7 & 3.6 via tox (HPC-7603)

boegel:
Both `Jenkinsfile` & `tox.ini` are automatically generated via `python -m vsc.install.ci -f` (cfr. https://github.com/hpcugent/vsc-install/pull/114)

Tests with Python 3.6 are allowed to fail for now in Travis (and will fail in Jenkins because `python3.6` is missing, for now).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,16 @@
-language: python
-python:
-  - "2.7"
-install:
-  - easy_install vsc-install
+dist: bionic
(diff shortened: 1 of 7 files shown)


boegel:
@stdweird please review...

stdweird:
@boegel looks ok, but is missing a version bump

PR #287, opened 2020-01-03
fix running tests on top of Python 3 (HPC-7603)

boegel:
With these changes, the full `vsc-base` test suite passes with both Python 2 and 3...

I don't consider this 100% done, I should review the changes again now that the full test suite passes and also take a look at test coverage, but I'd appreciate feedback on this already as is...

One thing to be considered is introducing a new `vsc.utils.py23` module that provides some functionality to make code compatible with both Python 2 and 3 in cleaner way (avoiding Python version checks as much as possible).


lib/vsc/utils/asyncprocess.py
```diff
@@ -179,18 +180,26 @@ def recv_some(p, t=.1, e=False, tr=5, stderr=False, maxread=None):
...
+    if sys.version_info[0] >= 3:
+        data = data.encode()
+
     while len(data):
         sent = p.send(data)
         if sent is None:
             raise Exception(MESSAGE)
         allsent += sent
-        data = buffer(data, sent)
+        if sys.version_info[0] >= 3:
```
stdweird: `is_py3` function must be part of py23 module; i'm not sure if there an idiom to allow easy switching but this is awful

lib/vsc/utils/fancylogger.py
```diff
@@ -186,12 +187,23 @@ def _env_to_boolean(varname, default=False):
...
 # register QUIET, EXCEPTION and FATAL alias
-logging._levelNames['EXCEPTION'] = logging.ERROR
-logging._levelNames['FATAL'] = logging.CRITICAL
-logging._levelNames['QUIET'] = logging.WARNING
+LOG_LEVEL_ALIASES = {
+    'EXCEPTION': logging.ERROR,
+    'FATAL': logging.CRITICAL,
+    'QUIET': logging.WARNING,
+}
+if sys.version_info[0] < 3:
```
stdweird: only the name of the dict should be part of the if/else, the for loop is identical
```python
if is_py3():
    levelnames = logging._nameToLevel
else:
    levelnames = logging._levelNames

for alias in LOG_LEVEL_ALIASES:
    levelnames[alias] = LOG_LEVEL_ALIASES[alias]
```

lib/vsc/utils/fancylogger.py
```diff
@@ -215,11 +227,11 @@ class MissingLevelName(KeyError):
 
 def getLevelInt(level_name):
     """Given a level name, return the int value"""
-    if not isinstance(level_name, basestring):
+    if not isinstance(level_name, str):
```
stdweird: euhm, that is not ok, right? also no point in hiding the "lack of basestring" mess in python3, make a py23 module and fix this

lib/vsc/utils/fancylogger.py
```diff
@@ -744,7 +751,7 @@ def setLogLevel(level):
     """
     Set a global log level for all FancyLoggers
     """
-    if isinstance(level, basestring):
+    if isinstance(level, str):
```
stdweird: nah, please fix properly

lib/vsc/utils/generaloption.py
```diff
@@ -29,25 +29,31 @@
...
-import StringIO
 import sys
 import textwrap
+from functools import reduce
 from optparse import OptionParser, OptionGroup, Option, Values
 from optparse import BadOptionError, SUPPRESS_USAGE, OptionValueError
 from optparse import SUPPRESS_HELP as nohelp  # supported in optparse of python v2.4
 from optparse import gettext as _gettext  # this is gettext.gettext normally
+
+try:
```
stdweird: no, please use a `is_py3` function or at the very least the sys.version crap

lib/vsc/utils/generaloption.py
```diff
@@ -29,25 +29,31 @@
...
+    import ConfigParser as configparser
+except ImportError:
+    from io import StringIO  # Python 3
+    import configparser
+
 from vsc.utils.dateandtime import date_parser, datetime_parser
 from vsc.utils.docs import mk_rst_table
 from vsc.utils.fancylogger import getLogger, setroot, setLogLevel, getDetailsLogLevels
-from vsc.utils.missing import shell_quote, nub
+from vsc.utils.missing import is_string, nub, shell_quote
```
stdweird: can we use the `is_string` for the fancylogger basestring replacements?

lib/vsc/utils/generaloption.py
```diff
@@ -545,7 +554,7 @@ def is_value_a_commandline_option(self, opt, value, index=None):
             return "Value '%s' starts with a '-'" % value
 
         if not self.ALLOW_TYPO_AS_VALUE:
-            possibilities = self._long_opt.keys() + self._short_opt.keys()
+            possibilities = list(self._long_opt.keys()) + list(self._short_opt.keys())
```
stdweird: you can't add the `.keys` before converting/forcing to a list?

lib/vsc/utils/generaloption.py
```diff
@@ -725,7 +734,8 @@ def format_option_rsthelp(self, formatter=None):
                 if opt.help is not nohelp:
                     values.append(['``%s``' % formatter.option_strings[opt], formatter.expand_default(opt)])
 
-            res.extend(mk_rst_table(titles, map(list, zip(*values))))
+            columns = [list(x) for x in zip(*values)]
```
stdweird: no list comprehension, just wrap with `list` (it's already used to wrap the tuples)

lib/vsc/utils/generaloption.py
```diff
@@ -1669,11 +1678,18 @@ def generate_cmd_line(self, ignore=None, add_default=None):
...
+                    # and just having __getitem__ is not sufficient to support slicing;
+                    # so, just try slicing, and ignore if it fails
+                    elif hasattr(opt_value, '__getslice__') or hasattr(opt_value, '__getitem__'):
+                        try:
+                            if action == 'add_first':
+                                opt_value = opt_value[:-len(default)]
+                            else:
+                                opt_value = opt_value[len(default):]
+                        except TypeError:
+                            # if slicing fails, a TypeError "is not subscriptable" will be raised
```
stdweird: log an error at the very least (or remove the try/except). why would it be ok to simply ignore this?

lib/vsc/utils/missing.py
```diff
@@ -40,14 +40,15 @@
 @author: Stijn De Weirdt (Ghent University)
 """
 import shlex
+import sys
 import time
+from collections import namedtuple, Mapping
+from functools import reduce
 try:
```
stdweird: use `is_py3`

lib/vsc/utils/missing.py
```diff
@@ -57,7 +58,12 @@
 
 def is_string(item):
```
stdweird: should be moved to th epy23 module magic

lib/vsc/utils/optcomplete.py
```diff
@@ -164,7 +164,7 @@ def __call__(self, **kwargs):
         if self.CALL_ARGS_OPTIONAL is not None:
             all_args.extend(self.CALL_ARGS_OPTIONAL)
 
-        for arg in kwargs.keys():
+        for arg in list(kwargs.keys()):
```
stdweird: why is the `list()` needed?

lib/vsc/utils/optcomplete.py
```diff
@@ -105,7 +105,7 @@
 from optparse import OptionParser, Option
```
stdweird: i'm ok with dropping this module, it's unused because too much latency

lib/vsc/utils/rest.py
```diff
@@ -35,13 +35,20 @@
...
 """
 import base64
-import urllib
-import urllib2
 try:
     import json
 except ImportError:
     import simplejson as json
 
+try:
```
stdweird: use `is_py3`

lib/vsc/utils/rest.py
```diff
@@ -35,13 +35,20 @@
 @author: Jens Timmerman
 """
 import base64
-import urllib
-import urllib2
 try:
```
stdweird: remove the try/except, simplejson is from py24 days

lib/vsc/utils/run.py
```diff
@@ -899,7 +899,7 @@ def _loop_process_output(self, output):
 
         # qa first and then qa_reg
         nr_qa = len(self.qa)
-        for idx, (question, answers) in enumerate(self.qa.items() + self.qa_reg.items()):
+        for idx, (question, answers) in enumerate(list(self.qa.items()) + list(self.qa_reg.items())):
```
stdweird: oh boy, why couldn't they implement a add function
also, this is (well, already was) unsorted?

test/dateandtime.py
```diff
@@ -84,34 +86,34 @@ def suite():
     """
 #    # date_parser
```
stdweird: just remove all the commented prints

stdweird: or better yet, remove everyting after `def suite` above

test/fancylogger.py
```diff
@@ -30,16 +30,23 @@
...
 import re
 import sys
 import shutil
-
-from StringIO import StringIO
 import tempfile
+from random import randint
 from unittest import TestLoader, main, TestSuite
+
+try:
```
stdweird: use `is_py3` or move this to the py23 module

test/fancylogger.py
```diff
@@ -117,7 +123,7 @@ def _reset_fancylogger(self):
 
         # delete all fancyloggers
         loggers = logging.Logger.manager.loggerDict
-        for name, lgr in loggers.items():
+        for name, lgr in list(loggers.items()):
```
stdweird: why `list`?

test/fancylogger.py
```diff
@@ -217,7 +240,7 @@ def test_uft8_decoding(self):
             logger.info(msg)
             logger.warning(msg)
             logger.warn(msg)
-            if isinstance(msg, unicode):
+            if not isinstance(msg, str):
```
stdweird: `is_string`?

test/fancylogger.py
```diff
@@ -264,9 +287,15 @@ def test_log_callback(msg, cache=callback_cache):
         open(self.logfn, 'w').write('')
 
         # test handling of non-UTF8 chars
-        msg = MSG + u"\x81"
-        msgre_tpl_error = r"DEPRECATED\s*\(since v%s\).*\xc2\x81" % max_ver
-        msgre_warning = re.compile(r"WARNING.*Deprecated.* will no longer work in v%s:.*\xc2\x81" % max_ver)
+        msg = MSG + u'\x81'
+        if sys.version_info[0] >= 3:
```
stdweird: is_py3

test/rest.py
```diff
@@ -29,7 +29,13 @@
 @author: Jens Timmerman (Ghent University)
 """
 import os
-from urllib2 import HTTPError
+
+try:
+    # Python 3
+    from urllib.request import HTTPError
```
stdweird: add it to the py23 module, we can use this in other places too

stdweird:
minor nagging about py23. at the very least, add a `is_py3` function to `vsc.utils.missing`

lib/vsc/utils/asyncprocess.py
itkovian: I would argue we should also be aware of potential differences in the various py3 versions. Better to have some server range or some such?

lib/vsc/utils/generaloption.py
boegel: I'll make `mk_rst_table` make sure the arguments are lists rather than generator object, since that's where we're using those values multiple times, to prevent running into surprises if you re-use `mk_rst_table` in other places...

lib/vsc/utils/run.py
boegel: Yeah, these should really be sorted (but individually, otherwise entries in `self.qa_reg` could get precedence over `self.qa` items, which may break stuff... Will fix.

test/fancylogger.py
boegel: Because `.items()` returns a generator, and we're changing `loggers` while we're looping over it, which means trouble in Python 3

lib/vsc/utils/py2vs3/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+
+def is_py_ver(maj_ver, min_ver=0):
+    """Check whether current Python version matches specified version specs."""
+
+    curr_ver = sys.version_info
+
+    lower_limit = (maj_ver, min_ver)
+    upper_limit = (maj_ver + 1, 0)
+
+    return lower_limit <= curr_ver < upper_limit
```
stdweird: what does this do?

lib/vsc/utils/asyncprocess.py
boegel: added `is_py3` to `vsc.utils.py2vs3`, along with `is_py_ver` which be used to check for `>= 3.6` or so if needed (via `is_py_ver(3, 6)`)

lib/vsc/utils/fancylogger.py
boegel: ok, fixed

lib/vsc/utils/fancylogger.py
boegel: fixed using `not is_string(level_name)`, where `is_string` is provided by `vsc.utils.py2vs3`

lib/vsc/utils/fancylogger.py
boegel: done with `is_string`

lib/vsc/utils/generaloption.py
boegel: done

lib/vsc/utils/generaloption.py
boegel: done

lib/vsc/utils/generaloption.py
boegel: no, because `.keys()` returns an iterator in Python 3, and you can't add iterators...

lib/vsc/utils/generaloption.py
boegel: I've removed the try/except. If we trigger an exception here, the code that uses `generaloption` is doing something weird/wrong, so it's OK to let this fail hard in that case

lib/vsc/utils/missing.py
boegel: fixed

lib/vsc/utils/missing.py
boegel: done

lib/vsc/utils/optcomplete.py
boegel: because we modify `kwargs` by deleting stuff from it while we're looping over it, which results in a crash in Python 3...

lib/vsc/utils/optcomplete.py
boegel: it was easy enough to fix the broken tests, so I left it in

if we really want to remove it, we can follow up on that...

lib/vsc/utils/rest.py
boegel: fixed

lib/vsc/utils/rest.py
boegel: fixed

lib/vsc/utils/run.py
boegel: sorted now (but still separately for `self.qa` and `self.qa_reg`)

test/dateandtime.py
boegel: done

test/fancylogger.py
boegel: done

test/fancylogger.py
boegel: fixed

test/fancylogger.py
boegel: fixed

test/rest.py
boegel: fixed

lib/vsc/utils/py2vs3/__init__.py
boegel: It checks with the current Python version is between the lower limit and upper limit.

for example `is_py_ver(3,7)` will return `True` if the current Python version is 3.7.4 (since `3.7 <= 3.7 < 3.8` 

```
>>> 3.7 <= 3.7 < 3.8
True
>>> 3.7 <= 3.8 < 3.8
False
>>> 3.7 <= 3.6 < 3.8
False
```

That works as expected because comparisons can be chained in Python, see https://docs.python.org/2/reference/expressions.html?highlight=comparisons#comparisons

boegel:
@stdweird This should be good to go now (approved by both Travis CI & `jenkins10`, tests passing in both Python 2 & Python 3).

PR #291, opened 2020-03-25
fix resource leak by calling close() after running command in subprocess in set_columns() function provided by generaloption

boegel:
`vsc-config` tests are failing with Python 3 because warnings like this are popping up in stderr:

```
/usr/lib64/python3.6/subprocess.py:786: ResourceWarning: subprocess 7342 is still running
```

It's not entirely clear where they're coming from, but it may be the `get_columns` function which is called every time a `GeneralOption` instance is created, since it is running a (short lived) subprocess without properly cleaning up afterwards...

### DIFF
--- a/lib/vsc/utils/generaloption.py
+++ b/lib/vsc/utils/generaloption.py
@@ -66,7 +66,9 @@ def set_columns(cols=None):
         stty = '/usr/bin/stty'
         if os.path.exists(stty):
             try:
-                cols = int(os.popen('%s size 2>/dev/null' % stty).read().strip().split(' ')[1])
+                proc = os.popen('%s size 2>/dev/null' % stty)
+                cols = int(proc.read().strip().split(' ')[1])
+                proc.close()
             except (AttributeError, IndexError, OSError, ValueError):
                 # do nothing
                 pass
--- a/lib/vsc/utils/generaloption.py
+++ b/lib/vsc/utils/generaloption.py
@@ -35,6 +35,7 @@ import inspect
 import operator
 import os
 import re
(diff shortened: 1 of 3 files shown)


wpoely86:
in python 3 you can query this directly?
https://docs.python.org/3/library/shutil.html#shutil.get_terminal_size

lib/vsc/utils/generaloption.py
```diff
@@ -63,10 +65,13 @@ def set_columns(cols=None):
         return
 
     if cols is None:
-        stty = '/usr/bin/stty'
-        if os.path.exists(stty):
+        if is_py_ver(3, 3):
```
itkovian: Is that "at least"?

boegel: `>=`

boegel: well actually is `>= 3.3` and `< 4`

itkovian: Since they are making heavy changes in 3.9 (afaik), would that comparison become a potential problem?

boegel: Who knows, but we'll deal with that when we start using Python 3.9.

For now, we're only testing with Python 3.6, as that's the Python we'll be using on RHEL8.

We can easily extend to additional Python versions in Travis/Jenkins, and deal with failing tests.

For now, let's try and keep things rolling with Python 3.6...

Also, `shutil.get_terminal_size` is still there in 3.9, see https://docs.python.org/3.9/library/shutil.html#shutil.get_terminal_size

PR #294, opened 2020-03-31
fix use of iteritems (for Python 3) + add collections(.abc).Mapping to py2vs3

wpoely86:


wpoely86:
@boegel have a look

lib/vsc/utils/frozendict.py
```diff
@@ -50,7 +52,7 @@ def __repr__(self):
 
     def __hash__(self):
         if self.__hash is None:
-            self.__hash = reduce(operator.xor, map(hash, self.iteritems()), 0)
+            self.__hash = reduce(operator.xor, map(hash, iteritems(self)), 0)
```
boegel: Why not just use `.items()` instead?

And we should try and enhance the tests to make them fail in Python 3 when `.iteritems()` is used...

wpoely86: Because items() is not the same in python2 and python3.

PR #295, opened 2020-04-01
fix resource warning in Python 3 for vsc.utils.run

Jan-PieterBaert:
to avoid errors like `lib/jobcli/utils.py:171: ResourceWarning: unclosed file <_io.BufferedReader name=5>` we need to close the stdout from the process.

### DIFF
--- a/lib/vsc/utils/run.py
+++ b/lib/vsc/utils/run.py
@@ -396,7 +396,7 @@ class Run(object):
 
     def _cleanup_process(self):
         """Cleanup any leftovers from the process"""
-        if self._process.stdout is not None:
+        if self._process.stdout is not None and not self._process.stdout.closed:
             try:
                 self._process.stdout.close()
             except OSError as err:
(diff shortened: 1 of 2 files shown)


lib/vsc/utils/run.py
```diff
@@ -396,7 +396,10 @@ def _wait_for_process(self):
 
     def _cleanup_process(self):
         """Cleanup any leftovers from the process"""
-        pass
+        try:
+            self._process.stdout.close()
+        except OSError:
+            self.log.raiseException("_cleanup_process: failed to close stdout of the process")
```
boegel, line 402: mention the original error here:

```python
        except OSError as err:
            self.log.raiseException("_cleanup_process: failed to close stdout of the process: %s" % err)
```

lib/vsc/utils/run.py
```diff
@@ -396,7 +396,10 @@ def _wait_for_process(self):
 
     def _cleanup_process(self):
         """Cleanup any leftovers from the process"""
-        pass
+        try:
+            self._process.stdout.close()
```
boegel, line 400: Do we need to check whether it's still open? What about `stderr`?

Jan-PieterBaert: I will add a check to see if it's still open and I haven't had any problems with `stderr` yet, but if you want I can do the same for `stderr`

PR #298, opened 2020-04-21
avoid trouble caused by Unicode characters in output of commands

### LINKED ISSUE #297: UnicodeDecodeError in RunNoShellAsyncLoop._read_process
Example:

```
2020-04-02 18:42:58,147 ERROR      RunNoShellAsyncLoop MainThread  _read_process: read failed
  File "/Volumes/work/vsc-base/lib/vsc/utils/run.py", line 678, in _read_process
    return out.decode()
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 782: ordinal not in range(128)
```

I'll look into reproducing this in a test, and fixing, looks like this was introduced in #287

boegel:
fixes #297

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ _coloredlogs_pkgs = [
 ]
 
 PACKAGE = {
-    'version': '3.0.7',
+    'version': '3.1.0',
     'author': [sdw, jt, ag, kh],
     'maintainer': [sdw, jt, ag, kh],
     # as long as 1.0.0 is not out, vsc-base should still provide vsc.fancylogger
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -44,7 +44,7 @@
 ]
 
 PACKAGE = {
-    'version': '3.0.6',
+    'version': '3.0.7',
```
itkovian, line 47: I think this warrants a minor bump, no?

boegel: I don't know, does it?

It's actually just a bug fix. But fine, bumped in 64fbefc

PR #303, opened 2020-05-19
Allow adjusting the environment of the subprocess.

itkovian:


test/run.py
```diff
@@ -463,3 +463,8 @@ def test_cmdlist(self):
         self.assertErrorRegex(ValueError, "Non-string item", CmdList, ['foo', None])
         self.assertErrorRegex(ValueError, "Non-string item", CmdList, ['foo', ['bar', 'baz']])
         self.assertErrorRegex(ValueError, "Found one or more spaces", CmdList, 'this has spaces', allow_spaces=False)
+
+    def test_env(self):
+        ec, output = run(cmd="echo $MYENVVAR", env = {"MYENVVAR": "something"})
```
stdweird, line 468: `run` does not use subshell, you need to run a bashscript that does this. there are tests that do that

itkovian: Hmm, it does pass on the env to Popen, no? through

~~~~
            self._process = self._process_module.Popen(self._shellcmd, **self._popen_named_args)
~~~~

which should now have the env.

stdweird: it does not matter, without subshell `$HUPPEL` is a literal string

PR #305, opened 2020-10-06
URL encode the domain and path of URLs

lexming:
This PR adds URL encoding of the domain and path of URLs passed to the rest api. Currently, it only URL encodes the query string of URLs. However, it is possible that the rest client receives non conformant URLs and those should be encoded as well to avoid unnecessary exceptions from `urllib`.

Specifically, this is an issue with group names in the VSC account page that have white spaces, for instance _leuvenall-fluent group_, as those names can be part of the path of URLs. The tool `sync_django_ldap` from `vsc-administration` fails to complete the LDAP sync due to this issue.

### DIFF
--- a/lib/vsc/utils/rest.py
+++ b/lib/vsc/utils/rest.py
@@ -42,7 +42,7 @@ from functools import partial
 from future.utils import iteritems
 
 from vsc.utils import fancylogger
-from vsc.utils.py2vs3 import HTTPSHandler, Request, build_opener, urlencode
+from vsc.utils.py2vs3 import HTTPSHandler, Request, build_opener, urlencode, urlquote
 
 
 class Client(object):
@@ -111,7 +111,7 @@ class Client(object):
         Do a http get request on the given url with given headers and parameters
         Parameters is a dictionary that will will be urlencoded
         """
-        url = self._append_slash_to(url) + self.urlencode(params)
+        url = urlquote(self._append_slash_to(url)) + self.urlencode(params)
         return self.request(self.GET, url, None, headers)
 
     def head(self, url, headers=None, **params):
@@ -119,7 +119,7 @@ class Client(object):
         Do a http head request on the given url with given headers and parameters
         Parameters is a dictionary that will will be urlencoded
         """
-        url = self._append_slash_to(url) + self.urlencode(params)
+        url = urlquote(self._append_slash_to(url)) + self.urlencode(params)
         return self.request(self.HEAD, url, None, headers)
 
     def delete(self, url, headers=None, body=None, **params):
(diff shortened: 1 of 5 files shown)


boegel:
@lexming Tests are failing in Jenkins with:

```
$ pip3 install --ignore-installed --user tox
...
PermissionError: [Errno 13] Permission denied: '/var/lib/jenkins/.local/lib'
script returned exit code 2
```

That's because Jenkins is using the `Jenkinsfile` from current `master` rather than the one updated in the PR, because you're not a member of the team that has write access to this repo (so you're on "untrusted" contributor)... :-/

boegel:
@lexming Can we cover this change in `test/rest.py`?

lexming:
Hold on this a moment, this fixes the issues but it might be for the wrong reasons.

lexming:
One issue in the VSC rest API has been already solved and another one is in the hands of `vsc-tech`. So we can proceed with this PR, the fix here still applies and the domain and path of the URL have to be URL encoded because there are entities in the data base with white spaces in their names (and potentially any other non-conformant character with URLs).

@boegel I thought about a way to test this, but it is not trivial. Current tests use the rest API of github, but the problem there is that there are no entries in github encoded with the standard `%` codes. It replaces everything in the URL with dashes. So creating a specific wiki page named "Test rest API", will be converted to "test-rest-api". Using the VSC account page is probably out of the table due to the authentication requirements. Do you have any other idea?

lib/vsc/utils/rest.py
```diff
@@ -111,47 +111,47 @@ def get(self, url, headers=None, **params):
         Do a http get request on the given url with given headers and parameters
         Parameters is a dictionary that will will be urlencoded
         """
-        url = self._append_slash_to(url) + self.urlencode(params)
+        url = urlquote(self._append_slash_to(url)) + self.urlencode(params)
```
stdweird, line 114: or do it in self.request instead of each method?

lib/vsc/utils/rest.py
```diff
@@ -111,47 +111,47 @@ def get(self, url, headers=None, **params):
         Do a http get request on the given url with given headers and parameters
         Parameters is a dictionary that will will be urlencoded
```
stdweird, line 112: add that the url itself cannot be urlencoded

stdweird:
@lexming use mock?

lib/vsc/utils/rest.py
stdweird: or is the urlquote of the urlencode not idempotent (i guess not...)

lib/vsc/utils/rest.py
stdweird: so this PR breaks all code that passes already encoded urls (to address this issue most likely)

lexming:
This issue will be solved in the interface of the VSC account page by not allowing non-conformant group names. Therefore, this PR is not needed.

PR #309, opened 2021-02-24
Drop travis, add gh action

wpoely86:
As travis will be shut down in the near future, replace it with a equivalent github action flow.


.github/workflows/unittest.yml
```diff
@@ -0,0 +1,24 @@
+name: run python tests
+on: [push, pull_request]
+
+jobs:
+  python_unittests:
+      runs-on: ubuntu-latest
+      strategy:
+        matrix:
+          python: [2.7, 3.6]
```
boegel, line 9: Why not test for a couple more Python 3 versions too?

```suggestion
          python: [2.7, 3.6, 3.7, 3.8, 3.9]
```

boegel (changes requested):

.github/workflows/unittest.yml
wpoely86: we don't test those anywhere else?

boegel: But it's trivial to test them here, so why not?

boegel (approved):

wpoely86:
I guess the result will only show up here after merging it. You can see it in my fork: https://github.com/wpoely86/vsc-base/runs/1971252893

PR #311, opened 2021-03-10
fix daemon utils for py3

kwaegema:


lib/vsc/utils/daemon.py
```diff
@@ -65,7 +65,7 @@ def daemonize(self):
         sys.stderr.flush()
         si = open(self.stdin, 'r')
         so = open(self.stdout, 'a+')
-        se = open(self.stderr, 'a+', 0)
+        se = open(self.stderr, 'ba+', 0)
```
stdweird, line 68: why is it not required for stdout?

boegel: Because `stdout` is not being opened unbuffered (that's what the `0` is about).

kwaegema: unbuffered (the 0) is only possible with binary, so its either b with 0 or not b and not 0

stdweird: again, why is it not the same?

boegel: It's not our code, and non-buffered is usually only done for `stderr`?

stdweird: yeah, whatever the reason, it seems to be some standard see https://pagure.io/python-daemon/blob/master/f/daemon/runner.py#_132

stdweird:
@kwaegema py2 tests still fail

kwaegema:
@stdweird Hmm, weird, I'm not sure where this is coming from..
```FAIL: test_timeout (test.run.TestRun)

  File "/var/lib/jenkins/jobs/hpcugent_github.com/jobs/vsc-base/branches/PR-311/workspace/test/run.py", line 295, in test_timeout

    do_test(False)

  File "/var/lib/jenkins/jobs/hpcugent_github.com/jobs/vsc-base/branches/PR-311/workspace/test/run.py", line 273, in do_test

AssertionError: True is not false : main depth=2 pid (pids [22371, 22369, 22368]) is killed by timeout
```
It doesn't seem related to my change and I can't reproduce it on my laptop,. Also python tests / python_unittests (2.7) (pull_request)  is ok

PR #315, opened 2021-05-19
feat: allow authenticated + ssl mail

itkovian:


lib/vsc/utils/mail.py
```diff
@@ -68,8 +69,25 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
 class VscMail(object):
     """Class providing functionality to send out mail."""
 
-    def __init__(self, mail_host=None):
-        self.mail_host = mail_host
+    def __init__(self, mail_host=None, auth=False, mail_config=None):
+
+        opts = {
+            "from": ("Sender", None, "store", None),
+            "smtp": ("Host:Port", None, "store", "exploud.ugent.be:587"),
```
stdweird, line 76: an ugent server as default in vsc-base?

lib/vsc/utils/mail.py
```diff
@@ -68,8 +69,25 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
...
+            "smtp-auth-password": ("Authentication password", None, "store", None),
+            "smtp-auth-user": ("Authentication user", None, "store", None),
+            "smtp-use-starttls": ("Should we start a TLS session", bool, "store_true", False),
+        }
+
+        self.auth = auth
+
+        if self.auth:
+            self.go = simple_option(opts, config_files=mail_config)
```
stdweird, line 85: i don't like this tbh. this should be called outside this class

lib/vsc/utils/mail.py
```diff
@@ -68,8 +69,25 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
...
+            (host, port) = self.go.options.smtp.split(":")
+            self.mail_host = host
```
stdweird, line 87: it's better to set these always, and default for host can be `''` and port `0`. it also simplifies the code below

lib/vsc/utils/mail.py
```diff
@@ -85,13 +103,20 @@ def _send(self,
         """
 
         try:
-            if self.mail_host:
-                logging.debug("Using %s as the mail host", self.mail_host)
-                s = smtplib.SMTP(self.mail_host)
+            if self.auth:
+                logging.debug("Using %s as the mail host with authenticatio", self.mail_host)
+                s = smtplib.SMTP(host=self.mail_host, port=self.mail_port)
```
stdweird, line 108: so this call can be put outside the if/else block when using default `''` and `0`

lib/vsc/utils/mail.py
```diff
@@ -85,13 +103,20 @@ def _send(self,
...
-                logging.debug("Using the default mail host")
-                s = smtplib.SMTP()
-                s.connect()
+                if self.mail_host:
+                    logging.debug("Using %s as the mail host", self.mail_host)
+                    s = smtplib.SMTP(self.mail_host)
+                else:
+                    logging.debug("Using the default mail host")
+                    s = smtplib.SMTP()
+                    s.connect()
```
stdweird, line 119: is the indent correct here? no connect if mail_host is configured?

lib/vsc/utils/mail.py
```diff
@@ -105,6 +130,10 @@ def _send(self,
                 raise
             except smtplib.SMTPDataError as err:
                 raise
+
+            if self.auth:
```
stdweird, line 134: what does this do? (and in particular, why is it only needed / allowed if auth is used?)

boegel:
Please make sure the changes you're making are covered by the tests.

lib/vsc/utils/mail.py
itkovian: Good point.

itkovian:
I thought a bit about it and I am not too happy with it either.

lib/vsc/utils/mail.py
itkovian: It was like this forever, but I'm not sure now.

lib/vsc/utils/mail.py
```diff
@@ -68,8 +67,45 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
 class VscMail(object):
     """Class providing functionality to send out mail."""
 
-    def __init__(self, mail_host=None):
-        self.mail_host = mail_host
+    def __init__(self, mail_host=None, mail_port=None, smtp_auth_user=None, smtp_auth_password=None, smtp_use_starttls=False):
```
stdweird, line 70: defaults for host and port can be  `''` and `0`

lib/vsc/utils/mail.py
```diff
@@ -68,8 +67,45 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
...
+        self.mail_host = host
+        self.mail_port = port
+        self.smtp_auth_user = smtp_auth_user
+        self.smtp_auth_password = smtp_auth_password
+        self.smtp_use_starttls = smtp_use_starttls
+
+    def _connect(self):
+
+        logging.debug("Using %s as the mail host with authenticatio", self.mail_host)
+        if self.mail_host:
```
stdweird, line 90: use the defaults

lib/vsc/utils/mail.py
```diff
@@ -68,8 +67,45 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
...
 
-    def __init__(self, mail_host=None):
-        self.mail_host = mail_host
+    def __init__(self, mail_host=None, mail_port=None, smtp_auth_user=None, smtp_auth_password=None, smtp_use_starttls=False):
+
+        # we must not have the port in the host as this will fail the certificate check if needed
+        (host, port) = (mail_host, mail_port)
+        if mail_host:
+            try:
+                (host, port_) = mail_host.split(":")
```
stdweird, line 76: why is the even allowed?

itkovian: why is the what?

stdweird: the format with a `:` in it

itkovian: Dunno, somebody passed `explout.ugent.be:587` to it at some point :p

lib/vsc/utils/mail.py
```diff
@@ -68,14 +67,57 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
...
+
+        if self.smtp_auth_user and self.smtp_auth_password:
+            s.login(user=self.smtp_auth_user, password=self.smtp_auth_password)
+
+        s.connect()
+
+        return s
 
     def _send(self,
-              mail_from,
```
stdweird, line 75: why was this needed? it looks horrible

lib/vsc/utils/mail.py
```diff
@@ -68,14 +67,57 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
...
+            s = smtplib.SMTP()
+
+        if self.smtp_use_starttls:
+            context = ssl.create_default_context()
+            s.starttls(context=context)
+
+        if self.smtp_auth_user and self.smtp_auth_password:
+            s.login(user=self.smtp_auth_user, password=self.smtp_auth_password)
+
+        s.connect()
```
stdweird, line 112: should check the code again to introduce `s.quit` again

lib/vsc/utils/mail.py
itkovian: ah no, my bad, the config has `set smtp=exploud.ugent.be:587`

lib/vsc/utils/mail.py
```diff
@@ -68,14 +67,49 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
...
-    def __init__(self, mail_host=None):
+    def __init__(
+        self,
+        mail_host='',
+        mail_port=0,
+        smtp_auth_user=None,
+        smtp_auth_password=None,
+        smtp_use_starttls=False):
+
+        # we must not have the port in the host as this will fail the certificate check if needed
```
stdweird, line 78: comment makes no sense anymore?

lib/vsc/utils/mail.py
```diff
@@ -68,14 +67,49 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
...
         self.mail_host = mail_host
+        self.mail_port = mail_port
+        self.smtp_auth_user = smtp_auth_user
+        self.smtp_auth_password = smtp_auth_password
+        self.smtp_use_starttls = smtp_use_starttls
+
+    def _connect(self):
+
```
stdweird, line 86: docstring

lib/vsc/utils/mail.py
stdweird: please use the defaults

lib/vsc/utils/mail.py
stdweird: how was this resolved?

lib/vsc/utils/mail.py
```diff
@@ -68,14 +67,57 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
...
+        if self.smtp_auth_user and self.smtp_auth_password:
+            s.login(user=self.smtp_auth_user, password=self.smtp_auth_password)
+            logging.debug("Authenticated")
+
+        s.connect()
+
+        return s
+
+    def _send(
```
wdpypere, line 115: can you put the function defenition on one line?

lib/vsc/utils/mail.py
```diff
@@ -105,27 +142,33 @@ def _send(self,
...
+            raise VscMailError(
+                mail_host=self.mail_host,
+                mail_to=mail_to,
+                mail_from=mail_from,
+                mail_subject=mail_subject,
+                err=err)
+        finally:
+            s.quit()
+
+    def sendTextMail(
```
wdpypere, line 165: same here?

lib/vsc/utils/mail.py
stdweird: ok i missed the finally block

lib/vsc/utils/mail.py
itkovian: I went through the code again, defaults are used when there is nothing provided, so I am not sure I can follow here.

stdweird: i meant to replace the whole if/else with
```python
logging.debug("Using %s as the mail host and %d as mail port", self.mail_host, self.mail_port)
s = smtplib.SMTP(host=self.mail_host, port=self.mail_port)
```

boegel:
Merged without version bump?

PR #316, opened 2021-05-26
add support for CC and BCC in sendTextMail and sendHTMLMail

smoors:


lib/vsc/utils/mail.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, mail_host=None, mail_to=None, mail_from=None, mail_subject=No
         """Initialisation.
 
         @type mail_host: string
-        @type mail_to: string
+        @type mail_to: string or list of strings
```
wpoely86, line 50: is this a breaking change or can it never have worked with a list?

lib/vsc/utils/mail.py
```diff
@@ -125,20 +125,26 @@ def sendTextMail(self,
                      mail_from,
                      reply_to,
                      mail_subject,
-                     message):
+                     message,
+                     cc=None,
+                     bcc=None):
```
wpoely86, line 130: can also add these to the html mail?

smoors: sure, was not sure if it's useful

boegel (approved):
Looks good to me at first sight, but I'd like to get a second opinion from someone who knows where this is used, w.r.t. the bugfix for `mail_to` taking a comma-separated list of addresses, since passing an actual list of strings can never have worked...

@wdpypere @itkovian: Can you take a look?

lib/vsc/utils/mail.py
```diff
@@ -125,20 +125,26 @@ def sendTextMail(self,
...
+        @type mail_to: string
         @type mail_from: string
         @type reply_to: string
         @type mail_subject: string
         @type message: string
+        @type cc: string
+        @type bcc: string
 
-        @param mail_to: a valid recipient email address
+        @param mail_to: a valid recipient email address or comma-separated string of email addresses
```
stdweird, line 141: remove the `a valid recipient email address or`

lib/vsc/utils/mail.py
```diff
@@ -125,20 +125,26 @@ def sendTextMail(self,
...
         @param mail_from: a valid sender email address.
         @param reply_to: a valid email address for the (potential) replies.
         @param mail_subject: the subject of the email.
         @param message: the body of the mail.
+        @param cc: a valid CC email address or comma-separated string of email addresses
```
stdweird, line 146: remove the part before `or`

lib/vsc/utils/mail.py
```diff
@@ -125,20 +125,26 @@ def sendTextMail(self,
...
+        @param bcc: a valid BCC email address or comma-separated string of email addresses
```
stdweird, line 147: idem

lib/vsc/utils/mail.py
```diff
@@ -125,20 +125,26 @@ def sendTextMail(self,
...
+                     bcc=None):
         """Send out the given message by mail to the given recipient(s).
 
-        @type mail_to: string or list of strings
+        @type mail_to: string
         @type mail_from: string
         @type reply_to: string
         @type mail_subject: string
         @type message: string
+        @type cc: string
```
stdweird, line 138: as this is new, please use correct list type.

stdweird:
i'm not a fan of more strings that are lists.
you can make the to_mail and esp the new cc and bcc lists.
only for to_mail you should support the (as of now) deprecated behaviour that a single email addr as a string is passed.

smoors:
@stdweird thanks for the review. I prefer lists as well, was mainly concerned about not breaking stuff at your side and consistency.

smoors:
ping?

lib/vsc/utils/mail.py
stdweird: rephrase as `list of strings (or string, but deprecated)` or something

lib/vsc/utils/mail.py
```diff
@@ -107,7 +108,7 @@ def _send(self, mail_from, mail_to, mail_subject, msg):
         """Actually send the mail.
 
         @type mail_from: string representing the sender.
-        @type mail_to: string representing the recipient.
+        @type mail_to: string or list of strings representing the recipient.
```
stdweird, line 111: same remark wrt deprecating string

stdweird:
@smoors minor remakrs. next time this code is touched, we will need to refactor inseatd of copy/pasting identical code

PR #318, opened 2021-09-13
Fix mail connect

itkovian:


boegel:
@itkovian Can you clarify how removing that `s.connect()` line fixes something? What does it fix? How did it work fine before with that in place, what changed that this needs to be trashed now?

itkovian:
As I understand the docs, the connect is called if you init with arguments specifying a mail server etc. If you call it without arguments, it will connect to localhost. Or try to. Which fails.

boegel:
No version bump?

PR #327, opened 2022-05-03
Remove creating double slash in url

matesaki:
Merging domain url and path url creates double slash
which causes on localhost error 404 (Page not found).

Signed-off-by: Martin Sakin <marty@inuits.eu>


lib/vsc/utils/rest.py
```diff
@@ -215,6 +215,8 @@ def get_connection(self, method, url, body, headers):
             sep = '/'
         else:
             sep = ''
+        if self.url.endswith('/') and url.startswith('/'):
```
stdweird, line 218: what url are you trying to parse here?

matesaki:
When I running scripts form vsc-ugent.git, then in get_connection() function I got:
self.url: `http://localhost:8000/api/`
sep: 
url: `/monitoring/time/1651612143/`
... and merging creates: `http://localhost:8000/api//monitoring/time/1651612143/`
This leads to error 404 (Page not found).
I'm running vsc-accountpage localy with DEBUG mode, so, maybe it is more sensitive.

lib/vsc/utils/rest.py
stdweird: @matesaki i'm mainly puzzled why the code above does work.
also add a unittest for this

matesaki: Condition above check if there is no slash before merging. If true, add one.
This new condition check if there are to many slashes. If true, remove one.

It may be confusing because there are two variables url, first Ends with '/' and second Starts with '/'.

Trying to make a test ...

matesaki: don't know how to test it

stdweird: i think the whole code block, incl the previous lines can become
```
url = url.lstrip('/')
if self.url.endswith('/'):
    sep = '/'
else:
    sep = ''
```

matesaki: yes, this is equivalent
(rebased)

PR #328, opened 2022-05-17
Censor X-Auth-Token from logs

lexming:
The X-Auth-Token is used by the REST API of OceanStor.

### DIFF
--- a/lib/vsc/utils/rest.py
+++ b/lib/vsc/utils/rest.py
@@ -169,6 +169,8 @@ class Client(object):
         # censor contents of 'Authorization' part of header, to avoid leaking tokens or passwords in logs
         headers_censored = copy.deepcopy(headers)
         headers_censored['Authorization'] = '<actual authorization header censored>'
+        if 'X-Auth-Token' in headers_censored:
+            headers_censored['X-Auth-Token'] = '<actual authorization header censored>'
 
         fancylogger.getLogger().debug('cli request: %s, %s, %s, %s', method, url, body, headers_censored)
 
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ _coloredlogs_pkgs = [
 ]
 
 PACKAGE = {
-    'version': '3.4.5',
+    'version': '3.4.6',
     'author': [sdw, jt, ag, kh],
     'maintainer': [sdw, jt, ag, kh],
     # as long as 1.0.0 is not out, vsc-base should still provide vsc.fancylogger


lib/vsc/utils/rest.py
```diff
@@ -169,8 +170,22 @@ def request(self, method, url, body, headers, content_type=None):
         # censor contents of 'Authorization' part of header, to avoid leaking tokens or passwords in logs
         headers_censored = copy.deepcopy(headers)
         headers_censored['Authorization'] = '<actual authorization header censored>'
```
stdweird, line 172: use teh same set intersection magic ?

lexming: good point, done in https://github.com/hpcugent/vsc-base/pull/328/commits/5b034b91ae32421db0771fd7cfd06850a34f212d

stdweird:
@lexming can you rebase on master and bump the version again?

lexming:
@stdweird this PR is ready, however the tests have been broken by https://github.com/hpcugent/vsc-base/pull/329 , can you fix them?

lib/vsc/utils/rest.py
```diff
@@ -168,9 +169,23 @@ def request(self, method, url, body, headers, content_type=None):
...
-        headers_censored['Authorization'] = '<actual authorization header censored>'
+        secret_items = ['Authorization', 'X-Auth-Token']
+        for secret in set(headers_censored).intersection(secret_items):
+            headers_censored[secret] = '<actual authorization header censored>'
+
+        if body and not is_string(body):
+            # censor contents of body to avoid leaking passwords
+            body_censored = copy.deepcopy(body)
+            secret_items = ['password']
+            for secret in set(body_censored).intersection(secret_items):
```
wpoely86, line 180: this is not safe enough:
```python
>>> set([{'a':'b'}])
TypeError: unhashable type: 'dict'
```

Maybe just wrap it in a 'try' and continue if it fails?

wpoely86, line 180: this is not safe enough:
```python
>>> set([{'a':'b'}])
TypeError: unhashable type: 'dict'
```

Maybe just wrap it in a 'try' and continue if it fails?

lexming: fixed in https://github.com/hpcugent/vsc-base/pull/328/commits/b7849d6521a6b89d10c897560d2b37ffab6cb6fc

lexming: fixed in https://github.com/hpcugent/vsc-base/pull/328/commits/b7849d6521a6b89d10c897560d2b37ffab6cb6fc

lib/vsc/utils/rest.py
```diff
@@ -168,9 +169,33 @@ def request(self, method, url, body, headers, content_type=None):
 
         # censor contents of 'Authorization' part of header, to avoid leaking tokens or passwords in logs
         headers_censored = copy.deepcopy(headers)
-        headers_censored['Authorization'] = '<actual authorization header censored>'
+        secret_items = ['Authorization', 'X-Auth-Token']
+        try:
```
stdweird, line 173: move this in method or function, and reuse also below

stdweird (approved):

PR #330, opened 2022-06-14
add MonoidConcat class

lexming:
This PR adds a variation of the `Monoid` class specifically optimized to concatenate lists.

`Monoid` is currently used by vsc-filesystems to concatenate lists of filesystem/quota attributes and it is painfully slow. To the point where some scripts such as `dquota` become unusable.

The issue is that the `reduce` method used in `Monoid` does not scale. The table below shows the time needed to inject into a `MonoidDict` an increasing amount of lists with the current mappend used in vsc-filesystems, some other alternative mappends and the new `MonoidConcat` class (without `reduce`):

1.  Monoid with current mappend used by vsc-filesystems:
    ```
    l = lambda xs, ys: xs + ys
    ```
2. Monoid with alternative mappend with `list.extend`:
    ```
    def l(xs, ys):
        new_xs = list(xs)
        new_xs.extend(ys)
        return new_xs
    ```
3. Monoid with alternative mappend with `list.append`:
    ```
    def l(xs, ys):
        zs = []
        list(map(zs.append, xs))
        list(map(zs.append, ys))
        return zs
    ```
4. Monoid with alternative mappend with `itertools.chain`:
     ```
     l = lambda xs, ys: list(itertools.chain(xs, ys))
     ```
5. MonoidConcat without `reduce`
    

| Nr of records | 1: a + b (s) | 2: extend (s) | 3: append (s) | 4: chain (s) | 5: MonoidConcat (s) |
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| 100  | 0.0227  | 0.0469 | 0.0512 | 0.0502 | 0.0155 |
| 1000  | 0.3881  | 0.4753 | 1.9757 | 0.9320 | 0.1381 |
| 10000  | 30.4001  | 29.3562 | 171.8019 | 63.3293 | 1.3504 |
| 20000  | 122.6388  | 116.9318 | -- | -- | 2.3829 |

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ _coloredlogs_pkgs = [
 ]
 
 PACKAGE = {
-    'version': '3.4.7',
+    'version': '3.5.0',
     'author': [sdw, jt, ag, kh],
     'maintainer': [sdw, jt, ag, kh],
     # as long as 1.0.0 is not out, vsc-base should still provide vsc.fancylogger


lexming:
Not sure what happened with Jenkins, but running `tox` on my side works fine:
https://gist.github.com/53be3bff4225c96598eb2ddcfe3889a7

lib/vsc/utils/missing.py
```diff
@@ -137,6 +137,23 @@ def star(self):
         """Return a new similar monoid."""
         return Monoid(self.null, self.mappend)
 
+class MonoidConcat(object):
```
stdweird, line 140: i don't understand the requirment/benefit for using monoids, but i think it is better that optimised `mappend` are simply supported in the `Monoid` class itself. eg. is you pass string `concat` as mappend, it uses this optimised way (by avodinh reduce)

lib/vsc/utils/missing.py
```diff
@@ -137,6 +137,23 @@ def star(self):
...
+    """Variation of the Monoid object optimized to concatenate lists"""
+
+    def concatenate(self, base_list, extra_list):
+        """Extend the base list with the items of the extra list"""
+        base_list.extend(extra_list)
+        return base_list
+
+    def __call__(self, *args):
+        """When the monoid is called, concatenate all given lists"""
+        base_list = args[0]
```
stdweird, line 150: this code actually extends the first arg, instead of returning a new list with all concatted lists

stdweird:
what is the full code for your timings? in particular why not use `list(itertools.chain(*list_of_all_lists))` instead of using monid at all?

lexming:
Superseded by https://github.com/hpcugent/vsc-filesystems/pull/84

PR #331, opened 2022-10-04
don't call connect if mail host is specified

smoors:
as explained in the smtplib docs at https://docs.python.org/3/library/smtplib.html#smtplib.SMTP.connect
"This method is automatically invoked by the constructor if a host is specified during instantiation."

~~the code in mail.py is written in such a way that the mail_host must always be specified, so `connect` is always called automatically.~~

calling `connect` manually (i.e. a second time) causes the process to hang without sending any mails.

### DIFF
--- a/lib/vsc/utils/mail.py
+++ b/lib/vsc/utils/mail.py
@@ -108,7 +108,7 @@ class VscMail(object):
 
         If provided, use authentication and TLS.
         """
-        logging.debug("Using mail host %s, mail port %d", self.mail_host, self.mail_port)
+        logging.debug("Using mail host %s, mail port %s", self.mail_host, self.mail_port)
         s = smtplib.SMTP(host=self.mail_host, port=self.mail_port)
 
         if self.smtp_use_starttls:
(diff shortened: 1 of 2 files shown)


lib/vsc/utils/mail.py
```diff
@@ -115,8 +115,6 @@ def _connect(self):
             context = ssl.create_default_context()
             s.starttls(context=context)
             logging.debug("Started TLS connection")
-        else:
-            s.connect()
```
itkovian, line 119: I distinctly recall this was needed at some point.

smoors: @itkovian it's indeed still needed if the mail host is not specified.
I was wrong thinking that it must always be specified.
fixed in https://github.com/hpcugent/vsc-base/pull/331/commits/1b9ff71f551fd877f7673fb4f99f5228e315ecf0

smoors:
@itkovian any idea why jenkins fails?

itkovian:
@smoors 
```
FAIL: test_timeout (test.run.TestRun)
  File "/var/lib/jenkins/jobs/hpcugent_github.com/jobs/vsc-base/branches/PR-331/workspace/test/run.py", line 295, in test_timeout
    do_test(False)
  File "/var/lib/jenkins/jobs/hpcugent_github.com/jobs/vsc-base/branches/PR-331/workspace/test/run.py", line 273, in do_test
AssertionError: True is not false : main depth=2 pid (pids [13852, 13850, 13849]) is killed by timeout
```

PR #332, opened 2022-10-10
use utf-8 for MIMEText

smoors:


lib/vsc/utils/mail.py
```diff
@@ -204,7 +204,7 @@ def sendTextMail(
 
         logging.info("Sending mail [%s] to %s.", mail_subject, mail_to)
 
-        msg = MIMEText(message)
+        msg = MIMEText(message, 'plain', 'utf-8')
```
wpoely86, line 207: can you add some unicode (something not in ascii) in a test to verify?

itkovian (approved):

PR #333, opened 2022-11-25
Use port 25 with starttls instead of 587 for mailing by default

wpoely86:
@itkovian Would this be okay for you?

For us port 587 is not open by default and we rely on 25 with `STARTTLS`.

### DIFF
--- a/lib/vsc/utils/mail.py
+++ b/lib/vsc/utils/mail.py
@@ -72,10 +72,10 @@ class VscMail(object):
     def __init__(
         self,
         mail_host='',
-        mail_port=587,
+        mail_port=25,
         smtp_auth_user=None,
         smtp_auth_password=None,
-        smtp_use_starttls=False,
+        smtp_use_starttls=True,
         mail_config=None):
         """
         - If there is a config file provided, its values take precedence over the arguments passed to __init__
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ _coloredlogs_pkgs = [
 ]
 
 PACKAGE = {
-    'version': '3.4.9',
+    'version': '3.4.10',
     'author': [sdw, jt, ag, kh],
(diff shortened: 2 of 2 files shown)


lib/vsc/utils/mail.py
```diff
@@ -72,10 +72,10 @@ class VscMail(object):
     def __init__(
         self,
         mail_host='',
-        mail_port=587,
+        mail_port=25,
```
stdweird, line 75: weird defaults (port 587 without starttls makes no sense) sure, but why change them? are you using it somewhere where you can't configure it?

wpoely86: yes, the scripts that mail in vsc-filesystems-quota only use the mail hostname from vsc-config. I could change it to be tuple in vsc-config `(hostname, port)`. But that is a bit more work and needs changes on multiple places.

In my opinion, use port 25 as the default makes more sense? I also don't really see good reason to use 587 as long as you use starttls?

stdweird: well, port 25 is not supposed to be used for client-mailserver, only mailserver-mailserver. 
it's best if we can somehow configue it properly (even enabling/disabling starttls)

stdweird: @wpoely86 perhaps allow the smtp option to indicate a filename with the mailconfig  if it starts with a `/`?

wpoely86: Actually upon better reading the code, we could just provide `server:25` and that will also work.

But I would suggest that we flip on starttls by default?

wpoely86: Alternatively we can add a default configuration file and gracefully handle it if it's not present?

wpoely86: Is 25 only server-server? Didn't know that. I thought the only difference was that 587 enforces TLS.

PR #334, opened 2022-11-29
drop py27, py37 and py38, add py3.10 in github actions

wpoely86:


wpoely86:
python 3.11 doesn't like the fancylogger tests. I'm not sure why it fails.

boegel:
@wpoely86 Any reason why you want to drop testing with 3.7 and 3.8?

wpoely86:
Why would we test those versions? 3.6 is RHEL8, 3.9 is RHEL9 and future.

wdpypere:
@wpoely86 did you try 3.10? looking at the changelog for 3.11 I can't immediately find what causes this (although a lot of work around async stuff was done)

wpoely86:
Yes, see https://github.com/hpcugent/vsc-base/actions/runs/3573457420/jobs/6007542646

python 3.10 works. I also couldn't find any change that explains it.

wdpypere:
I'll fix in #337

PR #337, opened 2023-02-21
python3 only

wdpypere:
- remove internal usage of py2vs3 module
 - remove colorized dependency
 - remove future module usage
 - happy new year
 - update tested python version
 - some resource decorators

needs some serious vetting


wdpypere:
it could use way more cleanup, but this is big enough as it is to start. I would actually also just remove the daemon code, as it is unused afaik.

wdpypere:
python 3.11 fails because Mapping moved places and I deleted the fix made in https://github.com/hpcugent/vsc-base/pull/335, as I do feel it does not belong there, as it is not an issue for py2v3. A new fix should be made.

wdpypere:
as @wpoely86 also found out in #334 , python 3.11 and fancylogger are no friends. I'll look into it.

wdpypere:
hmm, python 3.11 also does not like run_timeout

wdpypere:
for run_timeout, now we don't need python2 support it seems to me we should switch from `subprocess.popen` to `subprocess.run`, it is more high level and supports timeout out of the box.

wdpypere:
I would actually propose to add python 3.11 testing in a separate PR and have the discussion there.

.github/workflows/unittest.yml
```diff
@@ -6,7 +6,7 @@ jobs:
       runs-on: ubuntu-20.04
       strategy:
         matrix:
-          python: [2.7, 3.6, 3.7, 3.8, 3.9]
+          python: [3.6, 3.9, 3.11]
```
wpoely86, line 9: Can you also pull in the other changes from #334?

wdpypere: @wpoely86 sure, but what do they do?

wpoely86: Fix deprecation warnings and use the newest version of all actions. And add python 3.10.

wdpypere: I was mainly wondering what the "newest version of all actions" means and implies.

.github/workflows/unittest.yml
```diff
@@ -6,7 +6,7 @@ jobs:
       runs-on: ubuntu-20.04
```
stdweird, line 6: changes to this file should be generated with eg the `vsc.install.ci` tool. i propose to start with py3.6 and 3.9 only

lib/vsc/utils/fancylogger.py
```diff
@@ -343,7 +323,7 @@ def raiseException(self, message, exception=None, catch=False):
             exception = self.RAISE_EXCEPTION_CLASS
 
         self.RAISE_EXCEPTION_LOG_METHOD(fullmessage)
-        raise_with_traceback(exception(message))
+        raise(exception(message))
```
stdweird, line 326: why is this removed/changed?

.github/workflows/unittest.yml
wpoely86: You need to read the change notes for that :stuck_out_tongue: 

You currently have deprecation warnings. As seen in #334  it still works and does what was intended.

lib/vsc/utils/fancylogger.py
```diff
@@ -559,10 +539,7 @@ def logToScreen(enable=True, handler=None, name=None, stdout=False, colorize=Col
     by default, logToScreen will log to stderr; logging to stdout instead can be done
     by setting the 'stdout' parameter to True
 
-    The `colorize` parameter enables or disables log colorization using
-    ANSI terminal escape sequences, according to the values allowed
-    in the `colorize` parameter to function `_screenLogFormatterFactory`
-    (which see).
+    The `colorize` is deprecated and does nothing.
```
stdweird, line 542: add warning when still used like the one below

lib/vsc/utils/frozendict.py
```diff
@@ -22,13 +22,13 @@
...
 from functools import reduce
-from future.utils import iteritems
+try:
+    # py 3.10+
+    from collections import Mapping  # noqa
+except ImportError:
+    # < py 3.10
+    from collections.abc import Mapping  # noqa
 
-from vsc.utils.py2vs3 import Mapping
```
stdweird, line 27: move this back to some module to handle all this crap somewhere
this will happen more, and it's really ugly to copy/paste these blocks

lib/vsc/utils/frozendict.py
```diff
@@ -52,7 +52,7 @@ def __repr__(self):
 
     def __hash__(self):
         if self.__hash is None:
-            self.__hash = reduce(operator.xor, map(hash, iteritems(self)), 0)
+            self.__hash = reduce(operator.xor, map(hash, self.items()), 0)
```
stdweird, line 55: are we sure this is the same?

lib/vsc/utils/missing.py
```diff
@@ -44,10 +44,15 @@
 import time
 from collections import namedtuple
 from functools import reduce
+try:
```
stdweird, line 47: import from something else

setup.py
```diff
@@ -33,37 +33,23 @@
...
     'install_requires': [
         'vsc-install >= %s' % VSC_INSTALL_REQ_VERSION,
-        'future >= 0.16.0',
     ],
-    'extras_require': {
-        'coloredlogs': _coloredlogs_pkgs,
-    },
     'setup_requires': ['vsc-install >= %s' % VSC_INSTALL_REQ_VERSION],
-    'tests_require': ['prospector'] + _coloredlogs_pkgs,
+    'tests_require': ['prospector'],
```
stdweird, line 51: this is odd. this should all come from vsc-install

.github/workflows/unittest.yml
wdpypere: I agree, I'll make a ticket for vsc-install

lib/vsc/utils/frozendict.py
wdpypere: ah, I found out that since python 3.3 you can always do `from collections.abc import Mapping` so no more testing needed. I'll fix.

lib/vsc/utils/py2vs3/py2.py
```diff
@@ -41,6 +42,7 @@
 FileExistsErrorExc = OSError  # noqa
 FileNotFoundErrorExc = OSError  # noqa
 
+logging.warning("py2 module deprecated, please stop using it.")
```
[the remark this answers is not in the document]
wdpypere: I left it in because I did not remove the whole py2vs3 thing, I can however remove it, but it might break in other packages. I was thinking about removeing the whole py2vs3 in one go at a later stage, when more cleanups have happened.  I'll make it a raise.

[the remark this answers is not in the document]
wdpypere: hmm, I might indeed just remove it and raise the exception in ```py2vs3/__init__.py```

setup.py
wdpypere: I'll try and remove it

.github/workflows/unittest.yml
wdpypere: https://github.com/hpcugent/vsc-install/issues/214

lib/vsc/utils/fancylogger.py
wdpypere: fixed

lib/vsc/utils/frozendict.py
stdweird: even better

lib/vsc/utils/frozendict.py
wdpypere: I'll doublecheck

wdpypere:
```
>>> iteritems(thisdict)
dict_items([('brand', 'Ford'), ('model', 'Mustang'), ('year', 1964)])
>>> thisdict.items()
dict_items([('brand', 'Ford'), ('model', 'Mustang'), ('year', 1964)])
```

it does

wdpypere:
@stdweird remarks addressed

.github/workflows/unittest.yml
```diff
@@ -1,24 +1,28 @@
...
+      run: pip install tox
+    - name: add mandatory git remote
+      run: git remote add hpcugent https://github.com/hpcugent/vsc-base.git
+    - name: Run tox
+      run: tox -e py
+    strategy:
+      matrix:
+        python:
+        - 3.6
+        - 3.9
```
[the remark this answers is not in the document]
wdpypere: @itkovian these are the base pythons for RHEL8 and RHEL9. We know we'll have to fix stuff for 3.11, that's why it is not in here. if people use other pythons, they are welcome to open a PR and give a reasoning, but for now I know of no other case than 3.6 and 3.9 . (and 3.11 on fedora)

itkovian (approved):

lib/vsc/utils/fancylogger.py
```diff
@@ -343,7 +323,7 @@ def raiseException(self, message, exception=None, catch=False):
             exception = self.RAISE_EXCEPTION_CLASS
 
         self.RAISE_EXCEPTION_LOG_METHOD(fullmessage)
-        raise_with_traceback(exception(message))
+        raise(exception(message)).with_traceback(tb)
```
stdweird, line 326: is this correct? `with_traceback` looks an exception method

so probably somethin g like
```
excp = exception(message)
if tb is not None:
    excp = excp.with_traceback(tb)
raise(excp)
```

test/fancylogger.py
```diff
@@ -661,27 +630,27 @@ class ScreenLogFormatterFactoryTest(TestCase):
 
     def test_colorize_never(self):
```
stdweird, line 631: why do we keep the colorize tests? they are not supposed to do anything anymore, right?

test/run.py
```diff
@@ -107,8 +106,8 @@ def test_startpath(self):
 
         # command was actually executed, in specified directory
         self.assertTrue(os.path.exists(os.path.join(self.tempdir, 'bar')))
-        txt = open(os.path.join(self.tempdir, 'bar')).read()
-        self.assertEqual(txt, 'foo\n')
+        with open(os.path.join(self.tempdir, 'bar')) as txt:
```
stdweird, line 109: use fp instead of txt

lib/vsc/utils/fancylogger.py
wdpypere: well spotted, I'll fix.

wdpypere:
@stdweird remarks addressed

lib/vsc/utils/fancylogger.py
```diff
@@ -323,7 +323,9 @@ def raiseException(self, message, exception=None, catch=False):
             exception = self.RAISE_EXCEPTION_CLASS
 
         self.RAISE_EXCEPTION_LOG_METHOD(fullmessage)
-        raise(exception(message)).with_traceback(tb)
+        if tb is not None:
+            raise exception(message).with_traceback(tb)
```
stdweird, line 327: why the copy/paste? use a variable and build it up

wdpypere: it's a variable less. preference I guess. I'll change.

wdpypere:
@stdweird remark fixed again :smile:

PR #340, opened 2023-03-09
handle e.g. empty dict body data properly

stdweird:


lib/vsc/utils/rest.py
```diff
@@ -172,18 +172,20 @@ def request(self, method, url, body, headers, content_type=None):
...
-        fancylogger.getLogger().debug('cli request: %s, %s, %s, %s', method, url, body_censored, headers_censored)
+        body_censored = body
+        if body is not None:
+            if isinstance(body, str):
+                # assume serialized bodies are already clear of secrets
+                logging.debug("Request with pre-serialized body, will not censor secrets")
+                body_censored = body
+            else:
+                # censor contents of body to avoid leaking passwords
+                secret_items = ['password']
```
wpoely86, line 183: This is already defined on L172?

wpoely86: Nevermind, misread

lib/vsc/utils/rest.py
```diff
@@ -172,18 +173,20 @@ def request(self, method, url, body, headers, content_type=None):
...
-            fancylogger.getLogger().debug("Request with pre-serialized body, will not censor secrets")
-            body_censored = body
-
-        fancylogger.getLogger().debug('cli request: %s, %s, %s, %s', method, url, body_censored, headers_censored)
+        body_censored = body
+        if body is not None:
+            if isinstance(body, str):
+                # assume serialized bodies are already clear of secrets
+                logging.debug("Request with pre-serialized body, will not censor secrets")
+                body_censored = body
```
wpoely86, line 181: This is whole `if` still needed if you already set it on L176?

lib/vsc/utils/rest.py
stdweird: it reuses a variable name, but a bit odd to read indeed

lib/vsc/utils/rest.py
stdweird: what do you mean?

wpoely86: before the `if` you already do `body_censored = body`, so except for the debug line, this branch does nothing?

stdweird: well, the log in relevant, but i indeed forgot to remove this line

wpoely86 (approved):
Good for me but I have no merge rights.

wdpypere:
@stdweird I reintroduced the old pylint hack to vsc-install, this should fix the test issues you were running into.

PR #343, opened 2023-10-23
use shutils terminal size instead of guessing ourselves

wdpypere:
removes our own subprocess stty guessing, available since python 3.3


stdweird:
@wdpypere is this tested?

wdpypere:
also, it's tested in the tests: https://github.com/hpcugent/vsc-base/blob/d3fd4287e8b8b3505bf140c7e41eb14bfd6c41a9/test/generaloption.py#L935

PR #344, opened 2023-10-25
p39: disable empty logging tests

wdpypere:
It seems that around python 3.7 log stream handling was changed. (can't find the actual change in the changelog)

 I believe this is a good thing as now "logging with logging" no longer disappears when using fancylogger, which used to happen. 
 
 We used to test specifically for this behaviour. How big the behaviour change is, I can't say. So I don't know if we should rework fancylogger or just disable the test. As I would like to get rid of fancylogger I would propose to just disable the test, as this "new" behaviour might make migrating away from facylogger easier.

but those are just my 50 cents.

### DIFF
--- a/test/fancylogger.py
+++ b/test/fancylogger.py
@@ -538,7 +538,7 @@ class FancyLoggerTest(TestCase):
             self.assertEqual(stringfile.getvalue(), '',
                              msg="logging.debug reports nothing when fancylogger loglevel is debug")
         else:
-            self.assertEqual(stringfile.getvalue(), f'DEBUG:root:{msg}:',
+            self.assertEqual(stringfile.getvalue(), f'DEBUG:root:{msg}\n',
                              msg="logging.debug reports something when fancylogger loglevel is debug")
 
         fancylogger.setroot()
--- a/test/fancylogger.py
+++ b/test/fancylogger.py
@@ -532,8 +532,14 @@ class FancyLoggerTest(TestCase):
         msg = 'this is my string'
         logging.debug(msg)
 
-        self.assertEqual(stringfile.getvalue(), '',
+        if sys.version_info < (3,7):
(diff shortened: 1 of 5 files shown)


wdpypere:
@stdweird this is not urgent at all, but at some point in the future I would like your opinion on this.

test/fancylogger.py
```diff
@@ -532,8 +532,10 @@ def test_zzz_fancylogger_as_rootlogger_logging(self):
         msg = 'this is my string'
         logging.debug(msg)
 
-        # fails on python 3.11
-        self.assertEqual(stringfile.getvalue(), '',
+        if sys.version_info < (3,7):
```
stdweird, line 535: implement the `else` block instead of just skipping it

wdpypere:
@stdweird remark addresses

test/fancylogger.py
wdpypere: done

test/fancylogger.py
```diff
@@ -532,9 +532,14 @@ def test_zzz_fancylogger_as_rootlogger_logging(self):
         msg = 'this is my string'
         logging.debug(msg)
 
-        # fails on python 3.11
-        self.assertEqual(stringfile.getvalue(), '',
-                         msg="logging.debug reports nothing when fancylogger loglevel is debug")
+        if sys.version_info < (3,7):
+            # logging stream handling was changed in python 3.7
+            # this is a good thing as messages can no longer go missing
+            self.assertEqual(stringfile.getvalue(), '',
```
stdweird, line 538: factor out the common parts. the expected value is the only difference?

wdpypere: that, and the message, I'll fix it

wdpypere:
@stdweird fixed

PR #346, opened 2023-11-10
cleanup more code

wdpypere:


lib/vsc/utils/docs.py
```diff
@@ -57,7 +57,7 @@ def mk_rst_table(titles, columns):
         width = max(map(len, columns[i] + [title]))
 
         # make line template
-        tmpl.append('{%s:{c}<%s}' % (i, width))
+        tmpl.append('{{{}:{{c}}<{}}}'.format(i, width))
```
stdweird, line 60: what is this unreadable crap?

lib/vsc/utils/py2vs3/py3.py
```diff
@@ -36,12 +36,7 @@
 from tempfile import TemporaryDirectory  # noqa
 from urllib.parse import urlencode, unquote  # noqa
 from urllib.request import HTTPError, HTTPSHandler, Request, build_opener, urlopen  # noqa
-try:
-    # py 3.10+
-    from collections import Mapping  # noqa
-except ImportError:
-    # < py 3.10
-    from collections.abc import Mapping  # noqa
+from collections.abc import Mapping # noqa
```
stdweird, line 39: so the comment was wrong?

test/exceptions.py
```diff
@@ -84,18 +82,17 @@ def test_loggedexception_specifiedlogger(self):
...
 
         # if logger is specified, it should be used
         logToFile(tmplog, enable=True)
         self.assertErrorRegex(LoggedException, 'BOOM', raise_loggedexception, 'BOOM', logger=logger1)
         logToFile(tmplog, enable=False)
 
         rootlog = getRootLoggerName()
-        log_re = re.compile("^%s.testlogger_one :: BOOM( \(at .*:[0-9]+ in raise_loggedexception\))?$" % rootlog, re.M)
-        with open(tmplog, 'r') as f:
+        log_re = re.compile(f"^{rootlog}.testlogger_one :: BOOM( \\(at .*:[0-9]+ in raise_loggedexception\\))?$", re.M)
```
stdweird, line 92: use `rf"..` ?

test/exceptions.py
```diff
@@ -107,18 +104,17 @@ def test_loggedexception_callerlogger(self):
...
 
         # if no logger is specified, logger available in calling context should be used
         logToFile(tmplog, enable=True)
         self.assertErrorRegex(LoggedException, 'BOOM', raise_loggedexception, 'BOOM')
         logToFile(tmplog, enable=False)
 
         rootlog = getRootLoggerName()
-        log_re = re.compile("^%s(.testlogger_local)? :: BOOM( \(at .*:[0-9]+ in raise_loggedexception\))?$" % rootlog)
-        with open(tmplog, 'r') as f:
+        log_re = re.compile(f"^{rootlog}(.testlogger_local)? :: BOOM( \\(at .*:[0-9]+ in raise_loggedexception\\))?$")
```
stdweird, line 114: idem, `rf`?

test/py2vs3.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
stdweird, line 1: using the origina module should raise an exception at some point, or at least do a logging.error
and i think almost all these test can be simplified if not removed

lib/vsc/utils/py2vs3/py3.py
wdpypere: it seems so. I can't get it to break. Or it was fixed upstream.

```
[wdpypere@lancelot ~]$ python3.6
Python 3.6.15 (default, Aug  2 2023, 00:00:00) 
[GCC 13.2.1 20230728 (Red Hat 13.2.1-1)] on linux
>>> from collections.abc import Mapping
>>> 

[wdpypere@lancelot ~]$ python3.9
Python 3.9.18 (main, Aug 28 2023, 00:00:00) 
[GCC 13.2.1 20230728 (Red Hat 13.2.1-1)] on linux
>>> from collections.abc import Mapping
>>> 

[wdpypere@lancelot ~]$ python3.12
Python 3.12.0 (main, Oct  2 2023, 00:00:00) [GCC 13.2.1 20230918 (Red Hat 13.2.1-3)] on linux
>>> from collections.abc import Mapping
>>> 
```

boegel: https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping says "New in version 3.5"

lib/vsc/utils/docs.py
wdpypere: to escape a `{` you need to double it.... I'll see if I can come up with some more readable.

wdpypere: rewritten with ljust builtin instead of format on format on ...

test/py2vs3.py
wdpypere: trying to use py2 will already give an import error. I'll add a logging statement when you try to load the py3 version. I think I cleaned most of the code. We could also remove it, but things might break.

stdweird: up to you

wdpypere: warning added

wdpypere:
@stdweird remarks addressed

PR #348, opened 2023-11-17
remove py2vs3 module

wdpypere:
should only be done when:
 - https://github.ugent.be/hpcugent/vsc-quattor/pull/99
 - https://github.ugent.be/hpcugent/vsc-cloud/pull/35
 - https://github.ugent.be/hpcugent/vsc-storage/pull/23

are merged and deployed


wdpypere:
closes #351

wdpypere:
@stdweird this should fix the tests. effectively goes back to the behaviour from before november 2023. Maybe we hit a bug in python which got reverted/fixed?

wdpypere:
was introduced in https://github.com/hpcugent/vsc-base/commit/4bb9c770f8292848cd69fd907c75bddff1e5f4e6

PR #362, opened 2026-02-24
fix: use the epel version for tenacity on el9

itkovian:


setup.py
```diff
@@ -37,15 +37,15 @@
 from vsc.install import shared_setup
 from vsc.install.shared_setup import ag, jt, kh, sdw, wdp
 
-TENACITY_VERSION_LIMIT = "== 9.1.2"
+TENACITY_VERSION_LIMIT = "== 8.2.3"
 
 if sys.version_info >= (3, 10):
     TENACITY_VERSION_LIMIT = ">= 9.1.2, <= 9.2.0"
```
stdweird: is there really an upperbound?

setup.py
```diff
@@ -37,15 +37,15 @@
 from vsc.install import shared_setup
 from vsc.install.shared_setup import ag, jt, kh, sdw, wdp
 
-TENACITY_VERSION_LIMIT = "== 9.1.2"
+TENACITY_VERSION_LIMIT = "== 8.2.3"
 
 if sys.version_info >= (3, 10):
-    TENACITY_VERSION_LIMIT = ">= 9.1.2, <= 9.2.0"
+    TENACITY_VERSION_LIMIT = ">= 8.2.3"
```
stdweird: what is the purpose of these limitations? it really does not work with more recent versions?

setup.py
```diff
@@ -37,22 +37,19 @@
...
     "author": [sdw, jt, ag, kh],
     "maintainer": [sdw, jt, ag, kh, wdp],
     "install_requires": [
-        "vsc-install >= 0.17.19",
-        f"tenacity {TENACITY_VERSION_LIMIT}",
+        "vsc-install >= 0.24.1",
     ],
 }
 
+if sys.version_info < (3, 7):  # noqa: UP036
```
stdweird: this is worse. injecting deps this late gives no overview.
add the dependency without version in the normal PACKAGES dick, and append the version limit string here?

itkovian: FFS make up yr minds. one person says A, another says not A.

setup.py
itkovian: - Python 3.9 is supported up to 9.1.2
- Python 3.6 is supported up to 8.2.2
- EPEL offers 8.2.3 only

setup.py
```diff
@@ -37,22 +37,17 @@
...
     "author": [sdw, jt, ag, kh],
     "maintainer": [sdw, jt, ag, kh, wdp],
     "install_requires": [
-        "vsc-install >= 0.17.19",
-        f"tenacity {TENACITY_VERSION_LIMIT}",
+        "vsc-install >= 0.24.1",
     ],
 }
 
+if sys.version_info >= (3, 7):  # noqa: UP036
```
stdweird, line 49: i am completly lost now. does work or not on older python versions? and is the imit required or not?

itkovian: There is not RPM at this moment for el8. I need to build one.

itkovian: Without upper bounds, tests will grab what they can, resulting in failing tests as well:

~~~
File "/home/runner/work/vsc-base/vsc-base/.eggs.py39/tenacity-9.1.4-py3.9.egg/tenacity/__init__.py", line 101, in <module>
    P = t.ParamSpec("P")
AttributeError: module 'typing' has no attribute 'ParamSpec'
~~~

stdweird: i would not bother with el8 rpms tbh.
you can version limit in the setup file and remove it when making the rpm, so it wil pick up whatever the repo provides and no need to always keep this in sync with the OS of the day. see https://github.ugent.be/hpcugent/vsc-quattor/blob/master/setup.py#L60

wdpypere: we need el8 rpms, we still have servers depending on this code

wdpypere: ah, I see the bypass

wdpypere: fine

stdweird: i meant no tenacity support on el8. ofcourse we need el8 vsc-base rpm

wdpypere (approved):

